PR #3, opened 2015-09-16
Fix pip version changes

tobyoxborrow:
This isn't urgent. I think it works without this change. Just a bit tidier. Most important thing is it stops pip getting downgraded which may be an issue later.

In a fresh vagrant (vagrant destroy -f ; vagrant up) you would see the
following messages. Though the pip packages install, pip actually ends
up downgraded.

    ==> dev: % (cmd_desc, proc.returncode))
    ==> dev: OSError
    ==> dev: :
    ==> dev: Command /vagrant/.venv/bin/python -c "import sys, pip;
             sys...d\"] + sys.argv[1:]))" setuptools pip failed with error
             code 1

pip downgrades itself after installing the sh module:

    ==> dev: Installing collected packages: pip
    ==> dev:   Found existing installation: pip 7.1.2
    ==> dev:     Uninstalling pip-7.1.2:
    ==> dev:       Successfully uninstalled pip-7.1.2
    ==> dev: Successfully installed pip-1.5.4
    ==> dev: You are using pip version 1.5.4, however version 7.1.2 is
             available.
    ==> dev: You should consider upgrading via the 'pip install
             --upgrade pip' command.
    ==> dev: You are using pip version 7.1.0, however version 7.1.2 is
             available.
    ==> dev: You should consider upgrading via the 'pip install
             --upgrade pip' command.

This seems to be caused by pip being upgraded via requirements.txt. So
this change upgrades pip outside of pip via easy_install.

The pip version we get via easy_install at time of writing is 7.1.2. It
runs fine and does not show the above messages.

Also removed python-pip from being installed via apt-get as
python-virtualenv comes with its own (albiet broken) and the version in
python-pip is quite old (1.5.x).


jorisroovers:
Awesome, thank you very very much sir! Merged.

PR #6, opened 2015-09-24
Ignore B6 when merge commit

tobyoxborrow:
B6 (body-is-missing) should be ignored when the commit is a merge commit.

Merge commits can be generated by automated systems like Gerrit and
those systems copy the title from the source commit but may not copy the
body. commit-message-6 gives an example.


jorisroovers:
Awesome, thanks!

PR #10, opened 2016-04-14
Python 2.6 support

jgrund:
Python 2.6 support

We have a project that uses Python 2.6 and would like 
to use this tool to verify our commit messages.

This PR imports `OrderedDict` from the `ordereddict` 
package if the built-in from `collections` is not present.

### DIFF
--- a/gitlint/config.py
+++ b/gitlint/config.py
@@ -7,7 +7,12 @@ try:
 except ImportError:  # pragma: no cover
     # python 3.x
     from configparser import ConfigParser, Error as ConfigParserError  # pragma: no cover
-from collections import OrderedDict
+try:
+    # python >= 2.7
+    from collections import OrderedDict
+except ImportError:  # pragma: no cover
+    # python 2.4-2.6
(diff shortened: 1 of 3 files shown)


jorisroovers:
Thanks for the PR :-)

It's interesting that you've uncovered a bug in gitlint in that it complains about merge commit titles exceeding the default min length of 72 chars. See here: https://travis-ci.org/jorisroovers/gitlint/jobs/123068718. I've created #11 for that and will fix this by adding an option for gitlint to ignore title length on merge commit.

Regarding your PR, can you also modify ```install_requires``` in ```setup.py``` to include the ordereddict dependency? That way your code will still work when installed from pypi.

I'll look into "officially" supporting 2.6 (i.e. enabling unit tests for 2.6 in travis) once your PR is merged :-)

jgrund:
Thanks! When do you expect the next version to be published?

jorisroovers:
Hopefully somewhere next week. Still a few other things I want to get in.

jorisroovers:
Released! Thanks again :-)

PR #12, opened 2016-06-15
Add --debug option

slipcon:
add --debug option and suppress the "Using config from ..." command line
log when --debug is not set.   This should make gitlint silent when the
comment passed to it is well-formed, but nicely verbose by default on
any errors.


jorisroovers:
Awesome, thanks for submitting this! Let me see if there is some other low hanging fruit that I can include  in the 0.7.1 release. I'll try to make time to release it over the weekend.

PR #13, opened 2016-06-15
fix inverted violation message

slipcon:
the violation message when a title did not match the expected regex was
inverted, e.g. it said that the title "does match" when it should have
been "does not match"


jorisroovers:
Awesome, thanks!

PR #21, opened 2017-03-05
Add support for linting a range of commits natively.

### LINKED ISSUE #14: Feature suggestion: Add support for running on all commits in a branch
One can in theory just pass each commit into gitlint in a shell loop or something, but since Python initialization is somewhat slow, it'd be significantly faster and cleaner if one could do e.g. `gitlint origin/master..HEAD` and have it lint all the commits in that series.

tommyip:
Fixes #14.

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -70,6 +70,7 @@ def build_config(ctx, target, config_path, c, extra_path, ignore, verbose, silen
 @click.option('-c', multiple=True,
               help="Config flags in format <rule>.<option>=<value> (e.g.: -c T1.line-length=80). " +
                    "Flag can be used multiple times to set multiple config values.")  # pylint: disable=bad-continuation
+@click.option('--commits', default="HEAD", help="The range of commits to lint. [default: HEAD]")
 @click.option('-e', '--extra-path', help="Path to a directory with extra user-defined rules",
               type=click.Path(exists=True, resolve_path=True, file_okay=False, readable=True))
 @click.option('--ignore', default="", help="Ignore rules (comma-separated by id or name).")
@@ -79,14 +80,14 @@ def build_config(ctx, target, config_path, c, extra_path, ignore, verbose, silen
 @click.option('-d', '--debug', help="Enable debugging output.", is_flag=True)
 @click.version_option(version=gitlint.__version__)
 @click.pass_context
-def cli(ctx, target, config, c, extra_path, ignore, verbose, silent, debug):
+def cli(ctx, target, config, c, commits, extra_path, ignore, verbose, silent, debug):
     """ Git lint tool, checks your git commit messages for styling issues """
 
     # Get the lint config from the commandline parameters and
     # store it in the context (click allows storing an arbitrary object in ctx.obj).
     config, config_builder = build_config(ctx, target, config, c, extra_path, ignore, verbose, silent, debug)
 
-    ctx.obj = (config, config_builder)
+    ctx.obj = (config, config_builder, commits)
 
     # If no subcommand is specified, then just lint
     if ctx.invoked_subcommand is None:
@@ -101,7 +102,7 @@ def lint(ctx):
     try:
         if sys.stdin.isatty():
             # If target has not been set explicitly before, fallback to the current directory
-            gitcontext = GitContext.from_local_repository(lint_config.target)
+            gitcontext = GitContext.from_local_repository(lint_config.target, ctx.obj[2])
         else:
             stdin_str = ustr(sys.stdin.read())
             gitcontext = GitContext.from_commit_msg(stdin_str)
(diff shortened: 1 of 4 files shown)


tommyip:
This bought the time to lint Zulip/Zulip down to 6 mins 43 sec on my 5400rpm harddisk.

- [ ] Reduce overhead of fetching commit detail individually.
- [ ] Add test

gitlint/cli.py
```diff
@@ -117,8 +118,16 @@ def lint(ctx):
 
     # Let's get linting!
     linter = GitLinter(lint_config)
-    violations = linter.lint(last_commit)
-    linter.print_violations(violations)
+    number_of_commits = len(gitcontext.commits)
+    for index, commit in enumerate(gitcontext.commits):
+        violations = linter.lint(commit)
+        if violations:
+            click.echo(u"{0}Commit {1}:".format(
```
jorisroovers: I think we want to add an extra check here where we only print "Commit x" in case more than 1 commit has been linted. This ensures backwards compatibility as well.

gitlint/git.py
```diff
@@ -54,10 +54,11 @@ class GitCommit(object):
         In the context of gitlint, only the git context and commit message are required.
     """
 
-    def __init__(self, context, message, date=None, author_name=None, author_email=None, parents=None,
+    def __init__(self, context, message, sha, date=None, author_name=None, author_email=None, parents=None,
```
jorisroovers: I think we should make ```sha``` an optional argument because we won't always have the sha available. For example, gitlint supports just piping arbitrary text into it for linting, in this case we don't have the sha (that's the reason why all these other parameters are also optional)

gitlint/git.py
```diff
@@ -129,18 +132,36 @@ def from_local_repository(repository_path):
...
-            commit_author_name = sh.git.log("-1", "--pretty=%aN", **sh_special_args)
-            commit_author_email = sh.git.log("-1", "--pretty=%aE", **sh_special_args)
-            # %aI -> ISO 8601-like format, while %aI is strict ISO 8601, it seems to be less widely supporte
-            commit_date_str = sh.git.log("-1", "--pretty=%ai", **sh_special_args)
-            commit_parents = sh.git.log("-1", "--pretty=%P", **sh_special_args).split(" ")
-            commit_is_merge_commit = len(commit_parents) > 1
-
-            # changed files in last commit
-            changed_files_str = sh.git("diff-tree", "--no-commit-id", "--name-only", "-r", "HEAD", **sh_special_args)
+            for sha in sha_list:
```
jorisroovers: I like this a lot. Don't know what I hadn't thought about using a single call to git with a single ```--pretty``` line in the past.

tommyip:
I fixed some of the tests but this is mysterious (left and right are the same):
```
E AssertionError: "An error occurred while executing 'git rev-list HEAD^...': b'fatal: Random git error'" does not match
"An error occurred while executing 'git rev-list HEAD^...': b'fatal: Random git error'"
```

The remaining failing tests are the integration tests where it have problems accessing the side effect of `git rev-list HEAD^...`:
```
self = An error occurred while executing '/usr/bin/git rev-list HEAD^...': b"fatal: ambiguous argument 'HEAD^...': unknown re... working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'"
```
How should I go about fixing this?

gitlint/tests/test_git.py
```diff
@@ -44,36 +43,33 @@ def git_log_side_effect(*args, **_kwargs):
...
-                          call('-1', '--pretty=%P', _cwd=u"fåke/path", _tty_out=False)]
-
-        self.assertListEqual(sh.git.log.mock_calls, expected_calls)
+        expected_calls = [
+            call("rev-list", "HEAD^...", **expected_sh_special_args),
+            call.log(sample_sha, "-1", "--pretty=%aN,%aE,%ai,%P%n%B", _cwd=u"fåke/path", _tty_out=False),
+            call('diff-tree', '--no-commit-id', '--name-only', '-r', sample_sha, **expected_sh_special_args)
+        ]
+        print(sh.git.mock_calls[1])
+        print(expected_calls[1])
```
jorisroovers: Please remove these 2 print statements :)

gitlint/tests/test_git.py
```diff
@@ -84,45 +80,42 @@ def git_log_side_effect(*args, **_kwargs):
...
-        sh.git.log.reset_mock()
+        sh.git.assert_called_once_with("rev-list", "HEAD^...",
+                                       _tty_out=False, _cwd=u"fåke/path")
+        sh.git.reset_mock()
         err = b"fatal: Random git error"
-        sh.git.log.side_effect = ErrorReturnCode("git log -1 --pretty=%B", b"", err)
+        sh.git.side_effect = ErrorReturnCode("git rev-list HEAD^...", b"", err)
 
-        expected_msg = u"An error occurred while executing 'git log -1 --pretty=%B': {0}".format(err)
+        expected_msg = u"An error occurred while executing 'git rev-list HEAD^...': {0}".format(err)
```
jorisroovers: You need to escape ```^``` in ```expected_msg``` because the check below is a regex check, not an exact match check. So ```expected_msg``` should be:
```python
expected_msg = u"An error occurred while executing 'git rev-list HEAD\^...': {0}".format(err)
```

tommyip:
Thanks for the help! I fixed all the tests and also took advantage of the `--max-count` argument to allow gitlint to check both individual and range of commits.

I will try to add new tests and docs after the merge of this PR. BTW are you going to do a small release for this feature so Zulip don't have to wait till the release of v0.9.0?

jorisroovers:
Thank you very much for this excellent contribution! We now just have to get the docs updated and some new unit and integration tests added for this new use-case. If you work on the testing, I can add the docs. There are a few other minor changes I'm making to gitlint, I'm planning to release 0.8.1 in the next week or so.

Thanks again for this, very much appreciated!

Issue #24, opened 2017-03-14
--commits doesn't take commit specific config into account


jorisroovers:
We currently only apply the commit-specific config of the last commit, not of every commit in the specified range. The following code block needs to be moved to the for-loop below it:
https://github.com/jorisroovers/gitlint/blob/master/gitlint/cli.py#L127-L131

CC: @tommyip

Issue #23, opened 2017-03-14
--commits doesn't correctly handle refspec


jorisroovers:
For example, ```gitlint --commits HEAD~2``` isn't correctly handled: it only lints the first commit in that range instead of all. This is because we are checking for the presence of dots in the refspec and if not present, add an additional```--max-count``` argument when invoking git, see here: https://github.com/jorisroovers/gitlint/blob/master/gitlint/git.py#L127-L131

I think instead we will most likely have to add an additional check for the number of commits in the repository to correctly handle the refspec format.

CC: @tommyip

PR #25, opened 2017-03-15
Handle empty commit list in --commits (#22)

### LINKED ISSUE #22: --commits option does not handle 0 commits case.
### Command
```
$ gitlint --commits upstream/master...HEAD
```
where `upstream/master` and `HEAD` is the same commit.

### Error
```
...
...
File "/home/tommy/Documents/gitlint/gitlint/cli.py", line 114, in lint
    last_commit = gitcontext.commits[-1]
IndexError: list index out of range
```

### Solution
Either just tell the user there is no commit in the range or lint the HEAD commit. I prefer the second option but some user might find that inconsistent with how Git's range work.

tommyip:
This fixes a "index out of range" exception when the specified git range for --commits is empty.

Fixes #22.

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -121,6 +121,12 @@ def lint(ctx):
         click.echo(ustr(e))
         ctx.exit(GIT_CONTEXT_ERROR_CODE)
 
+    number_of_commits = len(gitcontext.commits)
+
+    if number_of_commits == 0:
+        click.echo(u'No commits in range "{0}".'.format(ctx.obj[2]))
+        ctx.exit(0)
+
     config_builder = ctx.obj[1]
     last_commit = gitcontext.commits[-1]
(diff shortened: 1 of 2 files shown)


tommyip:
The OSX test doesn't seem to be starting.

jorisroovers:
Thanks :) 

> The OSX test doesn't seem to be starting.
Yeah, there is an issue with the OSX tests on Travis that I don't think has anything to do with gitlint. I've reached out to Travis but haven't heard back yet.

With this merge, I'm going to release v0.8.1. I would've liked to have more tests and fixes for some of the other issues (#23 and #24), but I likely won't have time for those in the remainder of the week. Since I want to unblock you on https://github.com/zulip/zulip/issues/1131, I'll just note that the ```--commits``` feature is still experimental in the v0.8.1 release notes.

PR #28, opened 2017-04-23
Accept either a file or directory for the --extra-path option (#26)

### LINKED ISSUE #26: Ability to specify user rules as a file instead of directory
Currently gitlint can only take a directory as argument for extra rules `gitlint --extra-path=<path>`, would it make sense to also support only filenames?

tommyip:
The --extra-path option now allows user defined rules to be either a file or directory.

Fixes #26.


jorisroovers:
I think we might want an extra test or 2, but I'll add those as part of some pre 0.8.2 release work I'm doing in a bit. Merging this.

PR #34, opened 2017-06-29
Add handle of git scissors cleanup and custom comment char

ron8mcr:
Fix cleaning up commit message when `git commit -v` command used.

Git adds special cutline (scissors) when using with `-v` arg. Lines after this one will not be included to commit

See [--verbose arg description](https://git-scm.com/docs/git-commit#git-commit---verbose) and [scissors clean up](https://git-scm.com/docs/git-commit#git-commit-scissors) description in docs for details

Also add handle of custom comment char [see `core.commentChar` setting](https://git-scm.com/docs/git-config#git-config-corecommentChar)

### DIFF
--- a/gitlint/tests/samples/commit_message/sample1
+++ b/gitlint/tests/samples/commit_message/sample1
@@ -7,3 +7,8 @@ This line has a trailing tab.
 # ------------------------ >8 ------------------------
 # Anything after this line should be cleaned up
 # this line appears on `git commit -v` command
+diff --git a/gitlint/tests/samples/commit_message/sample1 b/gitlint/tests/samples/commit_message/sample1
+index 82dbe7f..ae71a14 100644
+--- a/gitlint/tests/samples/commit_message/sample1
++++ b/gitlint/tests/samples/commit_message/sample1
+@@ -1 +1 @@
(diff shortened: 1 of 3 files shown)


gitlint/tests/samples/commit_message/sample1
```diff
@@ -4,3 +4,6 @@ This is the first line of the commit message body and it is meant to test a line
 This line has a tråiling space. 
 This line has a trailing tab.	
 # This is a cömmented  line
+# ------------------------ >8 ------------------------
```
jorisroovers: I think we should add some non-commented text after the cutline to properly test this use-case since commented lines are being ignored regardless.

jorisroovers:
Thank you very much for submitting this PR! Except for the minor comment I left in the code review, this looks really good. 

I will try to get a release out that includes these and some other changes within the next 2 weeks.

gitlint/tests/samples/commit_message/sample1
ron8mcr: Fixed

jorisroovers:
Merged! Thanks :) I'll be working a releasing this and some other patches soon.

PR #37, opened 2017-09-12
Prevent Commas in text fields from breaking git log printing

joshholl:
We have had a few users with names formatted as "Surname, GivenName" that would cause git lint to error. In order to remedy this, I modified the delimiter to be a nul byte (which should never occur) for the git log --pretty command.


jorisroovers:
This is great, I apologize for the delay. Merging this. I'm planning to release 0.9.0 soon so this will be included in that.

Issue #38, opened 2017-09-14
Usage with `pre-commit`


sobolevn:
Hi, thanks for this package!

I have a question: what is the best way to integrate this library with http://pre-commit.com/ ?
Thanks!

asottile:
The interface for pre-commit is to pass a single argument on the commandline that is the commit-msg filename.  It seems that gitlint doesn't have an interface which enables this -- I'll see if I can't craft a PR.

A workaround would be to use the interface that gitlint uses in its hook file, something like this:

```yaml
-   id: gitlint
    name: gitlint
    language: python
    entry: bash -c 'cat "$@" | python -m gitlint.cli' --
    stages: [commit-msg]
```

sobolevn:
@asottile thanks! That's almost the same I came up with. 

Take a look at it: https://github.com/wemake-services/wemake-django-template/blob/master/%7B%7B%20cookiecutter.project_name%20%7D%7D/.pre-commit-config.yaml

asottile:
@sobolevn the problem with that definition is it'll be linting the commit that already happened and you really want it to be checking the in-progress commit :)

PR #39, opened 2017-09-20
Add --msg-filename option

asottile:
The --msg-filename option enables the following:
- Easier checking in a commit-msg hook without requiring a `cat` process.
- Easy integration with http://pre-commit.com

This is related to #38

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -118,13 +118,20 @@ def stdin_has_data():
 @click.option('-e', '--extra-path', help="Path to a directory or python module with extra user-defined rules",
               type=click.Path(exists=True, resolve_path=True, readable=True))
 @click.option('--ignore', default="", help="Ignore rules (comma-separated by id or name).")
+@click.option(
+    '--msg-filename', type=click.File(),
+    help='Path to a file containing a commit-msg',
+)
 @click.option('-v', '--verbose', count=True, default=0,
               help="Verbosity, more v's for more verbose output (e.g.: -v, -vv, -vvv). [default: -vvv]", )
 @click.option('-s', '--silent', help="Silent mode (no output). Takes precedence over -v, -vv, -vvv.", is_flag=True)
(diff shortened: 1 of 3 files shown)


asottile:
@jorisroovers I've rebased this branch (just noticed the conflicts -- sorry!)

asottile:
@jorisroovers any feedback on this patch?

pbregener:
Can't you already do `gitlint < commit_message.txt` instead?

asottile:
@pbregener that would require a shell

jorisroovers:
Thanks for your interest in gitlint. I apologize it took me this look to get back to you - life and other projects get in the way sometimes :)

Definitely like the idea.
I think we could actually do something more generic here by calling the parameter ```--commit-message``` and supporting the ```@``` syntax like many other tools do. Like so:

```
gitlint --commit-message "This is my commit message"
# Handy if you want to read in an env variable
gitlint --commit-message "$MESSAGE"

gitlint --commit-message @/path/to/my/commit/messagefile.txt
```

Thoughts?

Let me know if you still have time to work on this, if not, I'll see if I can squeeze it in.

asottile:
I've literally never seen a tool accept an argument like that. It should really be two separate options (either `--commit-msg ... ` or `--msg-filename ...`)

The wacky @ syntax isn't compatible with #38 either :(

jorisroovers:
Here's one example: https://github.com/jakubroztocil/httpie#request-items

Admittedly, the syntax not widely popular - but I've seen it used elsewhere.

Anyways, I'm good with this then. I'll merge it in - I'll update the docs and integration tests later (appreciate any PRs related to that).

asottile:
I think I already added some tests for ya :)

jorisroovers:
*integration* tests in the https://github.com/jorisroovers/gitlint/tree/master/qa directory, but all good - I can do that later.

asottile:
oh, I see -- sorry missed that :)

PR #46, opened 2017-12-22
Exit silently when the specified commit range is empty

tommyip:
Reduce confusion on whether the lint check is successful/failed.


tommyip:
Python 3.6 build failed due to `\r\n` return characters.

jorisroovers:
Merging this in, I'll follow up with a commit to add the debug statement.

PR #48, opened 2018-01-10
Fix failing on existing Git comment char

torwald-sergesson:
As soon as _git function returns string if it was executed (not sh.RunningCommand) **get_commentchar**
functions fails with AttributeError (could not get exit_code).


torwald-sergesson:
I see one failing check but it fails only on 2.7 and this check is not related to my code.

VladimirKasatkin:
Hi!
Will this pull request merged? I wanted to try gitlint on our team, but I do not want to use my own fork. Thanks!

jorisroovers:
Not sure when this actually occurs, but the code itself can't hurt, so merging this.

FYI that I'm not immediately planning on rolling another release - so you'll need to install from git if you want your team to use the version with this fix.

jorisroovers:
Thanks for your interest!

jorisroovers:
Related issue: #70

PR #49, opened 2018-01-12
Fixes for Debian packaging

RykHawthorn:
Hi,

I would like to create a package for Debian, therefore I came across some problems that need these fixes. The detailed information about the fixes are in the respective commits.

Here is the related "intend to package" (ITP) bug: https://bugs.debian.org/886864

Best
Patrik

### DIFF
--- a/gitlint/tests/base.py
+++ b/gitlint/tests/base.py
@@ -1,6 +1,12 @@
 import logging
 import os
-import unittest2
+
+try:
+    # python 2.x
+    import unittest2 as unittest
+except ImportError:
+    # python 3.x
+    import unittest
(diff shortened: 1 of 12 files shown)


jorisroovers:
Thanks for submitting. This is great :)

Do you need a 0.9.1 release for you to move ahead with the deb packaging? I'd need to find some time to do that - might take a week or two - so if you can continue without it, that's probably easier.

RykHawthorn:
Thanks.

No, we apply these changes in the packaging process and remove them once the next release includes them. So there is no need to hurry things up as we're not blocked by it.

PR #50, opened 2018-01-16
Enforce English in test_target

bdrung:
The test_target test case calls git and expect a certain error message in English. If a different languages is use, git might produce a translated error message that does not match the expected error message. Therefore set the environment variable `LANGUAGE=C`.

### DIFF
--- a/gitlint/tests/test_cli.py
+++ b/gitlint/tests/test_cli.py
@@ -283,6 +283,7 @@ class CLITests(BaseTestCase):
     @patch('gitlint.cli.stdin_has_data', return_value=False)
     def test_target(self, _):
         """ Test for the --target option """
+        os.environ["LANGUAGE"] = "C"
         result = self.cli.invoke(cli.cli, ["--target", "/tmp"])
         # We expect gitlint to tell us that /tmp is not a git repo (this proves that it takes the target parameter
         # into account).


RykHawthorn:
@bdrung The tests passed, but there was an import-error at the end:

```
$ coveralls
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.3.6/bin/coveralls", line 11, in <module>
    sys.exit(wear())
  File "/home/travis/virtualenv/python3.3.6/lib/python3.3/site-packages/coveralls/__init__.py", line 79, in wear
    from coveralls.control import coveralls
  File "/home/travis/virtualenv/python3.3.6/lib/python3.3/site-packages/coveralls/control.py", line 1, in <module>
    from coverage.control import Coverage
ImportError: cannot import name Coverage

Done. Your build exited with 0.
```
I think travis had a little hickup, because I can't see how your change relates to the above error message.

@jorisroovers can you trigger a rebuild?

jorisroovers:
We should probably do this in the base test class so it affects all tests, but I'll merge this already

bdrung:
You could either do it in every tests where it matters or set it for all tests. The other environment variable that should be set is LC_ALL to use UTF-8 in the test (e.g. set `LC_ALL=UTF-8`). Otherwise the tests will fail, if LC_ALL is set on ANSI.

Issue #42, opened 2017-11-30
Problem with Jenkins CI due to "sys.stdin.isatty" check


omarkohl:
When using gitlint on Jenkins it doesn't work as it does locally simply by executing "gitlint" to check the latest commit.

```
gitlint
```

Since on Jenkins there is not tty present the command simply fails with several cryptical warnings (probably because stdin is empty). When using the `-d` option it tells me `DEBUG: gitlint.lint Linting commit [SHA UNKNOWN]` which was even more confusing.

After reading the code and re-reading the documentation I ended up sucessfully using this:

```
git log -1 --pretty=%B | gitlint
```

This is a little cumbersome and it seems unecessary to base the decision whether to perform `GitContext.from_local_repository` depending on if there is a TTY present or not.

Suggestions:

* Always use `from_local_repository` if there is a local repository AND (stdin is empty OR not tty). Use stdin in other cases.
* Alternatively or additionally document this more clearly. Maybe a doc section on how to use gitlint with CI tools.

jorisroovers:
So I think that depends on how you have your Jenkins configured. I've been doing it this way on Jenkins for years (on multiple different Jenkins installations) - so it can definitely work.

I do agree that we can probably be a bit smarter about TTY detection and make the [CI section in our docs](http://jorisroovers.github.io/gitlint/#using-gitlint-in-a-cicd-script) more prominent.

Marking this as a feature enhancement. I don't know when I'll get to this - it's likely to take a while (I don't have as much time to work on this anymore).

Thanks for your interest and submitting this issue!

omarkohl:
Could you explain how you do it in Jenkins? Because I'm not aware of doing anything _special_. Just a regular Freestyle project with an "Execute shell" Build step.

Concerning the documentation I did read the part you linked to and it was actually where I got the final solution from. But it doesn't explain that this is actually necessary when not using a TTY. I only understood that when reading the Python code.

The resolution of this ticket has very low priority for me because now it works (as described above). But it might be helpful as documentation in case someone else runs into the issue.

jorisroovers:
So I looked into this a bit more, the Jenkins's I've worked with always had jobs configured to run in containers with a TTY attached - which explains it working for me.

Anyways, I changed gitlint to be smarter on how it detects what to do, removing the TTY detection piece entirely (basically, it just checks whether anything is being passed to STDIN at startup, if not, it falls back to the local repo).

[...]

PR #52, opened 2018-01-17
Fix stdin_has_data()

bdrung:
When calling gitlint in a custom git commit hook on a gitlab server, stdin_has_data() will always return True regardless whether gitlint is called directly or text is piped into it. To reproduce you can call this
script:

```python
#!/usr/bin/python3

import sys
import stat
import os

if sys.stdin.isatty():
    print("Input comes from tty.")
else:
    print("Input doesn't come from tty.")

mode = os.fstat(sys.stdin.fileno()).st_mode
if stat.S_ISFIFO(mode):
    print("stdin is piped")
elif stat.S_ISREG(mode):
    print("stdin is redirected")
else:
    print("stdin is terminal")
```

On my server it prints:

```
Input doesn't come from tty.
stdin is terminal
```

stdin should only be read if the user uses a pipe or a redirect to pass data in.

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -3,7 +3,7 @@
 import logging
 import os
 import platform
-import select
+import stat
 import sys
 
 import click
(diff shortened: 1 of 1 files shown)


jorisroovers:
I got similar feedback in #42 after I thought I fixed it with the last release. I need to make some more time to play with this, because this is a fundamental part of gitlint. I want to understand and test it properly myself before merging a PR related to it.

I don't know when I have time, but I'm going to try and make some in the next few weeks since this is a fairly big issue. Appreciate your patience!

I think the following should work as a temp workaround: ```git log -1 --pretty=%B | gitlint```

bdrung:
My workaround is to use this patch. :wink: So there is no hurry on my side.

Some test cases you can use:
```
$ echo | gitlint
$ gitlint < textfile
$ gitlint
```

pbregener:
I'm having the same issues as described in #42 (but using GitLab CI instead of Jenkins). This MR seems like a good solution.

pbregener:
BTW: The workaround you described @jorisroovers (piping the commit message body into `gitlint`) doesn't work if you have rules regarding author names and/or email addresses. A new release with a fix for this would be highly appreciated! 😄

bdrung:
@pbregener Have you tested my proposed patch? Can you comment how you configured GitLab CI? I am using this patch with GitLab.

pbregener:
Yes. I am running `gitlint` in a pretty standard GitLab CI config using the Docker executor and the latest `gitlab-runner` (v10.4.0).

bdrung:
v10.4.0 does not contain the proposed patch from this pull request. Can you apply the patch from this pull request and try again? How do you invoke gitlint?

pbregener:
I applied the patch from this pull request to gitlint. I'm not sure what the gitlab-runner version has to do with this. I invoke gitlint from a bash script, which is linked in the project's `.gitlab-ci.yml`.

pbregener:
You mentioned that you use it in GitLab with a commit hook which will be an entirely different thing.

jorisroovers:
Closing this out, this got fixed in 5c933091a8db448049410df56a08813f223ce780. More context in #42.

PR #53, opened 2018-01-18
Change formatting

bdrung:
We want to use gitlint in our company and therefore the error output should be self explaining. The first commit prints the commit hash on every line (to make the output shorter) and the second adds 'line' to the line number, because that is not intuitive.

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -172,10 +172,8 @@ def lint(ctx):
         ctx.exit(0)
 
     general_config_builder = ctx.obj[1]
-    last_commit = gitcontext.commits[-1]
 
     # Let's get linting!
-    first_violation = True
     exit_code = 0
     for commit in gitcontext.commits:
         # Build a config_builder and linter taking into account the commit specific config (if any)
@@ -189,14 +187,7 @@ def lint(ctx):
         # exit code equals the total number of violations in all commits
         exit_code += len(violations)
         if violations:
-            # Display the commit hash & new lines intelligently
-            if number_of_commits > 1 and commit.sha:
-                linter.display.e(u"{0}Commit {1}:".format(
-                    "\n" if not first_violation or commit is last_commit else "",
-                    commit.sha[:10]
-                ))
-            linter.print_violations(violations)
-            first_violation = False
+            linter.print_violations(commit.sha if commit else None, violations)
 
     # cap actual max exit code because bash doesn't like exit codes larger than 255:
     # http://tldp.org/LDP/abs/html/exitcodes.html
--- a/gitlint/lint.py
+++ b/gitlint/lint.py
@@ -84,14 +84,15 @@ class GitLinter(object):
         violations.sort(key=lambda v: (-1 if v.line_nr is None else v.line_nr, v.rule_id))
         return violations
 
-    def print_violations(self, violations):
+    def print_violations(self, commit_sha, violations):
         """ Print a given set of violations to the standard error output """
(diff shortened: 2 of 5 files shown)


jorisroovers:
Thanks for your suggestion :-)

I think this is somewhat subjective.

I actually looked at other linters like pep8 and pylint to determine the current format. I prefer to not change this in a backwards incompatible way, because there's folks who parse gitlint's output in their CI scripts. This change might break that.

I'd like to suggest an alternative approach where we introduce a ```--violation-format``` parameter that takes a format string, for example:

```bash
gitlint --violation-format "line <line_nr>: <rule_id> <violation_message> extra text"
```

Where we use expose variables ```<commit_sha>, <line_nr>, <rule_id>, <violation_message>, <violation_content>```

This way, we cater to anyone's desired output format, without making an backwards incompatible change.

The different levels of verbosity ```-v, -vv, -vvv``` then just become default values for format strings:

```bash
-v -> <line_nr>: <ruleid>
-vv -> <linenr>: <ruleid> <violation_message>
-vvv -> <linenr>: <ruleid> <violation_message>: "<violation_content>"
```

I've actually implemented something like this before in a now dormant project: https://github.com/jorisroovers/yamlpal
Code: https://github.com/jorisroovers/yamlpal/blob/master/yamlpal/dumper.py#L61-L85

I don't think I've got time to do this myself in the near future, but always welcome contributions!

Thoughts?

bdrung:
The idea of an `--violation-format` parameter sounds great. It is flexible and supports scripting. One question is open: How do you want to handle the grouping? Example:
```
Commit 1234:
-: RULE_ID_1
2: RULE_ID_2
```
vs
```
Commit 1234: -: RULE_ID_1
Commit 1234: 2: RULE_ID_2
```

pbregener:
Great suggestions @jorisroovers and @bdrung!

jorisroovers:
I apologize it took me this look to get back to you - life and other projects get in the way sometimes :)

I thought of this as well - and didn't include it in my initial response because I didn't have a good answer to it.

I think there's 2 options:
1. Add another option ```--violation-commit-format``` (maybe not the best name?). Pragmatic, but doesn't scale if we need to add more formatters in the future.
2. Add 2 parameters to ```--violation-format``` (a key-value pair). The key then specifies what you're setting the format for and the value is the actual format. Like so:
```
--violation-format "violation-commit" "MY CUSTOM COMMIT: {sha} FOO BAR"
--violation-format "violation" "line {line_nr} ..."
# Note that I replaced the angular brackets with curlies for variable substition, I think that's a more widely used convention
```
If we do this, we might as well make it even more generic and call it ```--format```.

Only thing I'm not sure of is how that would map onto ```.gitlint``` config. Probably with a new section like so:
```
[general.format]
violation-commit=line <line_nr> ...
violation=MY CUSTOM COMMIT: <sha> FOO BAR
```

I don't think ```--format``` neccessarily needs to be supported as part of ```.gitlint``` from the start, but we should probably do follow-up commits to do so, so that behavior is consistent (all commandline flags can also be set in ```.gitlint```).

bdrung:
Alternative format for option two: Use equal sign for assignment:
```
--violation-format violation-commit="MY CUSTOM COMMIT: {sha} FOO BAR"
--violation-format violation="line {line_nr} ..."
```
or maybe a config switch:
```
--config violation-commit-format="MY CUSTOM COMMIT: {sha} FOO BAR"
--config violation-format="line {line_nr} ..."
```

PR #57, opened 2018-03-05
Add regex for commit title that skips body checks, if matches

SteffenKockel:
We needed a possibility to exclude release commits from our rigorous gitlint-checks.

This pull requests has the following to offer

* An option to specify a regular expression and, if it matches on the commit messages title, skip further body checks for the given commit.
* An option to set the loglevel from commandline. This behavior was splitted between `-d` and `-v[vvv]`

### DIFF
--- a/gitlint/files/gitlint
+++ b/gitlint/files/gitlint
@@ -44,6 +44,7 @@
 
 # [body-min-length]
 # min-length=5
+# ignore-body-checks-regex=''
 
 # [body-is-missing]
 # Whether to ignore this rule on merge commits (which typically only have a title)
--- a/gitlint/rules.py
+++ b/gitlint/rules.py
@@ -81,14 +81,14 @@ class RuleViolation(object):
         return equal
 
     def __str__(self):
-        return sstr(self)  # pragma: no cover
+        return self.__unicode__()  # pragma: no cover
 
     def __unicode__(self):
         return u"{0}: {1} {2}: \"{3}\"".format(self.line_nr, self.rule_id, self.message,
                                                self.content)  # pragma: no cover
 
     def __repr__(self):
-        return self.__str__()  # pragma: no cover
+        return self.__unicode__()  # pragma: no cover
 
 
 class MaxLineLength(LineRule):
--- a/gitlint/rules.py
+++ b/gitlint/rules.py
@@ -1,9 +1,11 @@
 import copy
+import logging
 import re
 
 from gitlint.options import IntOption, BoolOption, StrOption, ListOption
 from gitlint.utils import sstr
 
+LOG = logging.getLogger("gitlint")
 
 class Rule(object):
     """ Class representing gitlint rules. """
@@ -247,10 +249,19 @@ class BodyFirstLineEmpty(CommitRule):
 class BodyMinLength(CommitRule):
     name = "body-min-length"
     id = "B5"
-    options_spec = [IntOption('min-length', 20, "Minimum body length")]
+    options_spec = [IntOption('min-length', 20, "Minimum body length"),
+                    StrOption('ignore-body-checks-regex', '', 'Ignore body checks for tiles matching regex')]
+
 
     def validate(self, commit):
         min_length = self.options['min-length'].value
+        regex = self.options['ignore-body-checks-regex'].value
+
+        if regex and re.search(regex, commit.message.title):
+            logmsg = "Skipping %s (%s), because ignore-body-checks-regex matches." % (self.id, self.name)
+            LOG.info(logmsg)
+            return
+
         body_message_no_newline = "".join([line for line in commit.message.body if line is not None])
         actual_length = len(body_message_no_newline)
         if actual_length > 0 and actual_length < min_length:
@@ -261,12 +272,23 @@ class BodyMinLength(CommitRule):
 class BodyMissing(CommitRule):
     name = "body-is-missing"
     id = "B6"
-    options_spec = [BoolOption('ignore-merge-commits', True, "Ignore merge commits")]
+    # This needs to come from the real options
+    options_spec = [BoolOption('ignore-merge-commits', True, "Ignore merge commits"),
+                    StrOption('ignore-body-checks-regex', '', 'Ignore body checks for tiles matching regex')]
+
 
     def validate(self, commit):
+        regex = self.options['ignore-body-checks-regex'].value
+
         # ignore merges when option tells us to, which may have no body
         if self.options['ignore-merge-commits'].value and commit.is_merge_commit:
             return
+
+        if regex and re.search(regex, commit.message.title):
+            logmsg = "Skipping %s (%s), because ignore-body-checks-regex matches." % (self.id, self.name)
+            LOG.info(logmsg)
+            return
+
         if len(commit.message.body) < 2:
             return [RuleViolation(self.id, "Body message is missing", None, 3)]
 
--- a/gitlint/tests/test_body_rules.py
+++ b/gitlint/tests/test_body_rules.py
@@ -178,3 +178,20 @@ class BodyRuleTests(BaseTestCase):
         violations = rule.validate(commit)
         expected_violation_2 = rules.RuleViolation("B7", "Body does not mention changed file 'bar.txt'", None, 4)
         self.assertEqual([expected_violation_2, expected_violation], violations)
+
+    def test_skip_body_checks_regex(self):
+
+        # Skip body checks, if regex matches
+        rule = rules.BodyMinLength()
+        rules.BodyMinLength({'ignore-body-checks-regex': 'Release\s([a-z\-])*\s(\d+\.)?(\d+\.)?(\*|\d+)'})
+        commit = self.gitcommit(u"Release foo-bar-baz 0.12.3")
+        violations = rule.validate(commit)
+        self.assertIsNone(violations)
+
+        # Skip body checks, if regex matches
+        rule = rules.BodyFirstLineEmpty()
+        rules.BodyMinLength({'ignore-body-checks-regex': 'Release\s([a-z\-])*\s(\d+\.)?(\d+\.)?(\*|\d+)'})
+        commit = self.gitcommit(u"Release foo-bar-baz 0.12.3")
+        violations = rule.validate(commit)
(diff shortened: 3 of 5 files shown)


gitlint/files/gitlint
```diff
@@ -44,6 +44,8 @@
 
 # [body-min-length]
 # min-length=5
+# ignore-body-checks-regex=''
+
```
bdrung: unnecessary empty line

gitlint/rules.py
```diff
@@ -79,14 +81,14 @@ def __eq__(self, other):
         return equal
 
     def __str__(self):
-        return sstr(self)  # pragma: no cover
+        return self.__unicode__()  # pragma: no cover
```
bdrung: Why do you use `self.__unicode__()`? Does this work with Python 2 too?

gitlint/rules.py
```diff
@@ -243,14 +245,23 @@ def validate(self, commit):
             if first_line != "":
                 return [RuleViolation(self.id, "Second line is not empty", first_line, 2)]
 
-
```
bdrung: Please restore the newline. Otherwise pep8 will complain.

gitlint/rules.py
```diff
@@ -243,14 +245,23 @@ def validate(self, commit):
...
     id = "B5"
-    options_spec = [IntOption('min-length', 20, "Minimum body length")]
+    options_spec = [IntOption('min-length', 20, "Minimum body length"),
+                    StrOption('ignore-body-checks-regex', '', 'Ignore body checks for tiles matching regex')]
+
 
     def validate(self, commit):
         min_length = self.options['min-length'].value
+        regex = self.options['ignore-body-checks-regex'].value
+        regex_valid = True if (self.options['ignore-body-checks-regex'].value) else False
```
bdrung: regex_valid is not needed (see below).

gitlint/rules.py
```diff
@@ -243,14 +245,23 @@ def validate(self, commit):
...
+
+        if regex_valid and re.search(regex, commit.message.title):
```
bdrung: You can directly check for regex:
```python
         if regex and re.search(regex, commit.message.title): 
```

gitlint/rules.py
```diff
@@ -243,14 +245,23 @@ def validate(self, commit):
...
+            logmsg = "Skipping %s (%s), because ignore-body-checks-regex matches." % (self.id, self.name)
+            LOG.info(logmsg)
```
bdrung: Do not use `%`, but pass the formatting string to the logger (because that is faster if the log message does not be printed):
```python
LOG.info("Skipping %s (%s), because ignore-body-checks-regex matches.", self.id, self.name)
```

gitlint/rules.py
```diff
@@ -261,12 +272,24 @@ def validate(self, commit):
...
     id = "B6"
-    options_spec = [BoolOption('ignore-merge-commits', True, "Ignore merge commits")]
+    # This needs to come from the real options
+    options_spec = [BoolOption('ignore-merge-commits', True, "Ignore merge commits"),
+                    StrOption('ignore-body-checks-regex', '^Release', 'Ignore body checks for tiles matching regex')]
+
 
     def validate(self, commit):
+        regex = self.options['ignore-body-checks-regex'].value
+        regex_valid = True if (self.options['ignore-body-checks-regex'].value) else False
```
bdrung: Same as above.

gitlint/rules.py
```diff
@@ -261,12 +272,24 @@ def validate(self, commit):
 class BodyMissing(CommitRule):
     name = "body-is-missing"
     id = "B6"
-    options_spec = [BoolOption('ignore-merge-commits', True, "Ignore merge commits")]
+    # This needs to come from the real options
+    options_spec = [BoolOption('ignore-merge-commits', True, "Ignore merge commits"),
+                    StrOption('ignore-body-checks-regex', '^Release', 'Ignore body checks for tiles matching regex')]
```
bdrung: The default of `ignore-body-checks-regex` should be empty.

gitlint/rules.py
```diff
@@ -261,12 +272,24 @@ def validate(self, commit):
...
+        regex_valid = True if (self.options['ignore-body-checks-regex'].value) else False
+
         # ignore merges when option tells us to, which may have no body
         if self.options['ignore-merge-commits'].value and commit.is_merge_commit:
             return
+
+        if regex_valid and re.search(regex, commit.message.title):
+            logmsg = "Skipping %s (%s), because ignore-body-checks-regex matches." % (self.id, self.name)
+            LOG.info(logmsg)
+            return []
```
bdrung: To be consistent, just return nothing.

gitlint/tests/test_body_rules.py
```diff
@@ -77,12 +77,12 @@ def test_body_min_length(self):
         commit = self.gitcommit("Title\n\nThis is the second body line\n")
 
         violations = rule.validate(commit)
-        self.assertIsNone(violations)
+        self.assertIn(violations, [[],None])
```
bdrung: This change is not needed if you return None.

gitlint/tests/test_body_rules.py
```diff
@@ -77,12 +77,12 @@ def test_body_min_length(self):
...
 
         # assert no error - no body
         commit = self.gitcommit(u"Tïtle\n")
         violations = rule.validate(commit)
-        self.assertIsNone(violations)
+        self.assertIn(violations, [[], None])
```
bdrung: This change is not needed if you return None.

gitlint/rules.py
SteffenKockel: Yes, this works also with python 2 (Tested on python 2.7.14)

gitlint/cli.py
```diff
@@ -122,14 +122,15 @@ def stdin_has_data():
               help="Verbosity, more v's for more verbose output (e.g.: -v, -vv, -vvv). [default: -vvv]", )
 @click.option('-s', '--silent', help="Silent mode (no output). Takes precedence over -v, -vv, -vvv.", is_flag=True)
 @click.option('-d', '--debug', help="Enable debugging output.", is_flag=True)
```
bdrung: `--debug` should be either removed or mapped to `loglevel=debug`.

gitlint/tests/test_body_rules.py
```diff
@@ -70,6 +70,12 @@ def test_body_first_line_empty(self):
         violations = rule.validate(commit)
         self.assertListEqual(violations, [expected_violation])
 
+        # Skip body checks, if regex matches
+        rules.BodyMinLength({'ignore-body-checks-regex': 'Release\s([a-z\-])*\s(\d+\.)?(\d+\.)?(\*|\d+)'})
+        commit = self.gitcommit(u"Release foo-bar-baz 0.12.3")
+        violations = rule.validate(commit)
+        self.assertIsNone(violations)
+
```
bdrung: That test should go into a separate test method.

gitlint/tests/test_body_rules.py
```diff
@@ -111,6 +117,12 @@ def test_body_min_length(self):
         violations = rule.validate(commit)
         self.assertIsNone(violations)
 
+        # Skip body checks, if regex matches
+        rules.BodyMinLength({'ignore-body-checks-regex': 'Release\s([a-z\-])*\s(\d+\.)?(\d+\.)?(\*|\d+)'})
+        commit = self.gitcommit(u"Release foo-bar-baz 0.12.3")
+        violations = rule.validate(commit)
+        self.assertIsNone(violations)
+
```
bdrung: That test should go into a separate test method.

gitlint/tests/test_body_rules.py
SteffenKockel: Hmm, sounds like harassment, but Done

gitlint/tests/test_body_rules.py
SteffenKockel: See above

gitlint/cli.py
SteffenKockel: This breaks even more tests and is not what we want to achieve here. Rejected.

gitlint/tests/test_body_rules.py
bdrung: The benefit of separate test cases is that it easier to spot which part breaks and it prevents hiding failures (the test will abort on the first assertion error).

jorisroovers:
It doesn't entirely sit right with me is that this is being implemented as an option for specific rules.

IMO the solution should be generic.

More something along the lines of:

```
[ignore.title-match]
regex=^Release(.*)
ignore=T1,body-min-length,B6
# Note that like with general ignore, ignore accepts both names and ids

[ignore.body-line-match]
regex=^Hotfix
ignore=all

# This can be added later
[ignore.meta-author-match]
regex=bossman@company.com
ignore=all
```

There's several hours of work involved with implementing this I think (and like usually, the majority of time spent on tests and docs).

Interested to hear your thoughts.

Let me know if you intend on working on this, if not, I can probably look at it at some point (today, next month, later this year, really depends on when I find/make time).

jorisroovers:
Put some more thought into this, I think I've got a better solution.

We can introduce a new type of rule ```IgnoreRule``` that are ran before other rules and when they match, gitlint will stop running other rules for that particular commit.

The nice thing about this approach is that it's generic and that other users can easily extend it by building their own custom ```IgnoreRules``` - similarly to the existing functionality for building [user-defined rules](http://jorisroovers.github.io/gitlint/user_defined_rules/).

Gitlint would then just implement a couple of built-in ignore-rules that are set to ignore nothing by default, but can be configured like so:

```
[ignore-by-title]
regex="^Release(.*)"

[ignore-by-body-line]
regex="^Hotfix(.*)"
```

I'll work on this :-)

jorisroovers:
This was implemented as part of 7d447959e4220428d929c4aa1310675bb9e3bee1 and 52dc67e751dbbaa66c455063c6fe4f2f86d668d4.

Will be released fairly soon as part of v0.10.0. Closing this out.

SteffenKockel:
This was actually my first contact with this code. Thanks for incorporating the requested feature. No bad feelings here. :+1:

PR #58, opened 2018-03-07
Add ability to set loglevel via cli argument

SteffenKockel:
The loglevel switch is cryptic and disfunctional. This commit add a command-line option (-l, --loglevel) to manually set the loglevel from command line. The default level is still error. 

Signed-off-by: Steffen Kockel <steffen.kockel@profitbricks.com>


jorisroovers:
Thanks for your interest in gitlint!

I'd like to keep ```--debug``` but just add ```--loglevel``` on top of that, and have ```--debug``` just default to ```--loglevel DEBUG```. That will also take care of the failing integration test.

Could you add tests and docs for this feature itself too? Thanks!

SteffenKockel:
Hi, I realized relatively late, that this breaks tests and was not very confident, that you would pick this specific change. Since you signaling now, that you would like to see the feature, I will look into it again. (May take some time)

jorisroovers:
Well, I think it's more a nice-to-have. Note that gitlint only does DEBUG logging today, no INFO, WARNING, etc - so right now, there's actually no value in setting the value to anything but DEBUG.

There's no haste :-)

PR #64, opened 2018-04-09
Removed erroneous 'stages' line from .pre-commit-hooks.yaml

rob-orr:
Removed the `stages` line from `.pre-commit-hooks.yaml`, and made a slight re-ordering to conform to the `pre-commit` docs.  The `stages` line belongs in `.pre-commit-config.yaml` instead.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,5 +1,4 @@
 -   id: gitlint
     name: gitlint
-    language: python
     entry: gitlint --msg-filename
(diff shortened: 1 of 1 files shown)


jorisroovers:
I don't use pre-commit myself, but I believe you :-)

CC: @asottile

jorisroovers:
So this went out with the v0.10.0 release, do we need to roll this back?

asottile:
Probably :) it's usable as is, but would be better with the original contents

PR #65, opened 2018-04-18
Loosen dependency version requirements

pbregener:
Instead of requiring *one specific version* of dependencies, we now require *at least* that version.


jorisroovers:
> This turned out to be less trivial than anticipated. Closing for now.
:-)

I would've probably pushed back on this regardless. I've shot myself in the foot enough to know to always explicitly pin dependencies to a specific version. In my experience, making it more loose is just ticking time-bomb for one of your dependencies to make a breaking change that then requires a fire-drill to patch gitlint (usually at the worst possible time).

PR #69, opened 2018-06-21
Unpin dependencies

keimlink:
Pinning `install_requires` to an exact version causes users of a library a lot of problems.

Why projects should never do this has been explained in the [Python Packaging User Guide][1] and in a [blog post][2] by Donald Stufft.

This pull request changes just `install_requires` from an exact version to lower bounds. It does not change `extras_require`.

It also updates Click and arrow to the latest version in `requirements.txt`.

[1]: https://packaging.python.org/discussions/install-requires-vs-requirements/
[2]: https://caremad.io/posts/2013/07/setup-vs-requirement/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 setuptools
 wheel==0.24.0
-Click==6.6
+Click==6.7
 sh==1.11; sys_platform != 'win32' # sh is not supported on windows
-arrow==0.10.0
+arrow==0.12.1
 ordereddict==1.1; python_version < '2.7'
 importlib==1.0.3; python_version < '2.7'
--- a/setup.py
+++ b/setup.py
@@ -65,8 +65,8 @@ setup(
         "License :: OSI Approved :: MIT License"
     ],
     install_requires=[
-        'Click==6.6',
-        'arrow==0.10.0'
+        'Click>=6.6',
+        'arrow>=0.10.0'
     ],
     extras_require={
         ':python_version < "2.7"': [


keimlink:
The tests for Python 2.6 fail because Arrow is in the process of [removing support for Python 2.6](https://github.com/crsmithdev/arrow/issues/517#issuecomment-364983737). Unfortunately their classifiers still contain Python 2.6.

Their dependencies configuration [only supports Python 2.7](https://github.com/crsmithdev/arrow/commit/61f4808d5f6bb116b5952b90b9c838e81a26e8b9) and does not install the `backports.functools_lru_cache` package for Python 2.6.

So I if a dependency of gitlint drops support for Python 2.6, gitlint should also drop it. I think this is ok because

1. gitlint is issuing warnings to Python 2.6 users for some time now.
2. Python is no longer maintained and people should definitely upgrade to 2.7 or better >=3.5.

If you still want to support Python 2.6 `backports.functools_lru_cache` should be added to the dependencies for Python 2.

jorisroovers:
Thanks for your interest in gitlint :)

I most definitely agree with this _in theory_, but not in practice. I've been burnt multiple times in the past by a dependency releasing a new version that breaks my code (IIRC, I even had this happen with gitlint at some point). I understand I could set an upper bound so that dependencies don't get upgraded to major new versions, but the reality is that libraries don't always strictly follow semver. I've had multiple experiences with library APIs changing on minor and even patch releases.

Ultimately this means that if I don't set dependencies to a specific version, it's only a matter of time before ```pip install gitlint``` fails because of a dependency being updated. At that point, it becomes a firedrill for me to update gitlint to deal with that and release a new version of gitlint with the fix.

For a hobby project, that's not a commitment I'm willing to make. I prefer the peace-of-mind that a given release of gitlint will keep working.

Thanks for understanding, and obviously I'm open to being convinced otherwise!

bunder:
Joris, with pinned versions there are problems with gitlint packaging for distributions, for example Arch Linux. It has [Click 6.7](https://www.archlinux.org/packages/community/any/python-click/) and [arrow 0.12.1](https://www.archlinux.org/packages/community/any/python-arrow/). Running gitlint in this environment causes errors. How this problem could be solved?

jorisroovers:
@bunder By using [virtualenv](https://virtualenv.pypa.io/en/stable/)  :-)

bunder:
It's not a solution for Linux distributions, it's workaround. Packages should not normally include it's own copy of all dependencies.

jorisroovers:
Again, I can see that, but like I outlined before, this is not scalable for me as the maintainer of gitlint - and it will jeopardize overall stability of gitlint as a result (i.e. dependency updates breaking gitlint).

In this case, you can probably also use ```pip install --no-deps``` as as workaround.

keimlink:
I'm leaving for vacation, but I will provide an example to convince you when I'm back in August.

jorisroovers:
Closing this in favor of #133

Issue #20, opened 2017-01-18
Windows not supported


otakot:
When trying to install on windows 7 with python 2.7.11 - I got following error trace: 

```
Collecting gitlint
  Using cached gitlint-0.8.0-py2.py3-none-any.whl
Collecting sh==1.11 (from gitlint)
  Using cached sh-1.11.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "c:\users\admin\appdata\local\temp\pip-build-q_3wmy\sh\setup.py", line 4, in <module>
        import sh
      File "sh.py", line 37, in <module>
        support." % __version__)
    ImportError: sh 1.11 is currently only supported on linux and osx. please install pbs 0.110 (http://pypi.python.org/pypi/pbs) for windows support.
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in c:\users\admin\appdata\local\temp\pip-build-q_3wmy\sh\
```

Installation of pbs 0.110 did not help to resolve the issue - same error occurs

jorisroovers:
gitlint currently doesn't support Windows. No concrete plans either, might be something for the future.

otakot:
thanks

jorisroovers:
Keeping this open, as we might still consider this in the future.

Kristinita:
> gitlint currently doesn't support Windows. 

@jorisroovers , I think, it would be nice, if you will show it at the beginning of README and documentation, that Windows users do not waste their time.

Thanks.

jorisroovers:
Made some time to work on this a bit. Relevant commits: cc5e6d8da25ddbd9f3ac302f22d688dca38e62c4 and 20ac439ce94a1ba2628bb21f5e8621de398398e3.

I'm not a Windows user myself, but from the basic testing I did on a Windows 10 VM using Python 2.7 and Python 3.7, basic linting functionality seems to work. I'd appreciate anyone doing additional testing to confirm its working and report any bugs they find. I've already merged this work into master.

Things that work:
```sh
# lint latest commit
gitlint
gitlint -v --ignore T3,T5 # Options seem to work too

# Pass a message via stdin
echo "WIP: foöbar" > gitlint
git log -1 --pretty=%B | gitlint

# Commit ranges
gitlint --commits HEAD~2^..HEAD
```

However, I've also found a number of issues (some worse than others):
[...]

PR #71, opened 2018-07-11
Run on windows

byrney:
This MR updates gitlint to run on windows. The unit tests also run on windows but the qa tests don't.

* Replace use of sh with a (very) cut-down version in gitlint/sh.py (gitlint + unit tests)
* Update gitlint and unit tests to run on windows

The amount of sh functionality used by gitlint is quite small and neatly encapsulated inside the _git method. Originally I thought I could drop in a replacement for _git but nearly all the tests mock out sh so that would have required modifying almost all of them. Replacing sh meant I could keep the unit tests largely unchanged to support the shift to windows.

The changes for windows are straight-forward, mostly in the unit tests, slashes need to be `path.join` and the encoding when reading sample files needs to be explicity specified (the commit has more details). 

The QA tests use much more of the sh functionality (particularly fake ttys) so I left those as is for now. Some are easy enough to adapt (see this wip commit for examples: https://github.com/byrney/gitlint/commit/50bc2c93757b08f63ade2329f9a7af8a3b48756e) but others require significant effort or may never be runnable on windows.

If there is appetite I can do another PR to make the majority of QA run on both platforms and mark the rest to skip on windows.

Caveats

* I didn't update the docs. They say "windows is not supported" which may continue to be true. 
* Gitlint does not work in mintty (the git bash windows terminal) unless run under `winpty` (due to the way it detects terminal vs pipe)
* Calling it `sh.py` may be confusing since `sh` is still used in qa. The alternative is a big search/replace on the tests to use a new name for `gitlint/sh.py`
* I tested on py2.7 and py3.6 I'm hoping travis will tell me if there are problems with other versions when this PR gets CI'd.
* We are running this fork in our linux and CI envs now and will start rolling out to windows desktops this week/next

Related Issues: #20

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -14,13 +14,6 @@ USAGE_ERROR_CODE = 253  # noqa
 GIT_CONTEXT_ERROR_CODE = 254  # noqa
 CONFIG_ERROR_CODE = 255  # noqa
 
-# We need to make sure we're not on Windows before importing other gitlint modules, as some of these modules use sh
-# which will raise an exception when imported on Windows.
-if "windows" in platform.system().lower():  # noqa
-    click.echo("Gitlint currently does not support Windows. Check out "  # noqa
-               "https://github.com/jorisroovers/gitlint/issues/20 for details.", err=True)  # noqa
-    exit(USAGE_ERROR_CODE)  # noqa
-
 import gitlint
 from gitlint.lint import GitLinter
 from gitlint.config import LintConfigBuilder, LintConfigError, LintConfigGenerator
--- a/gitlint/config.py
+++ b/gitlint/config.py
@@ -9,6 +9,7 @@ import copy
 import re
 import os
 import shutil
+import codecs
 
 try:
     # python >= 2.7
@@ -313,7 +314,8 @@ class LintConfigBuilder(object):
         self._config_path = os.path.abspath(filename)
         try:
             parser = ConfigParser()
-            parser.read(filename)
+            with codecs.open(filename, encoding="UTF-8") as config_file:
+                parser.readfp(config_file, filename)  # pylint: disable=bad-option-value,deprecated-method
 
             for section_name in parser.sections():
                 for option_name, option_value in parser.items(section_name):
--- a/gitlint/git.py
+++ b/gitlint/git.py
@@ -1,5 +1,6 @@
 import sys
 from locale import getpreferredencoding
+import arrow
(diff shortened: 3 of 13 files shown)


jorisroovers:
Hi, 

this is awesome - thanks for doing this.

> The amount of sh functionality used by gitlint is quite small and neatly encapsulated inside the _git method

Yes! This was intentionally done a while back to make it easier for this change to be implemented at some point in the future - glad you found that :-)

I'm going to need some quality time to review and test this change myself. Given the foundational nature of this change, I'm inclined to be quite careful with accepting it without doing my own due diligence. I've been burnt before by the dozens of edge-cases that come with subprocess/shell execution (hence why I went to using ```sh``` originally).

In addition, I only tend to work on gitlint a couple of times a year, when I can make the time (this is a hobby project :-)). I promise I'll take a decent look at this at some point, but just a heads up that it might be a while (think months) before this actually get's merged and released.

Also, wrt Travis, I've see random integration test failures in the past - I'm fairly convinced this is a Travis issue because I've never had those issues outside of Travis and retriggering usually resolves these issues.

Thanks again for this PR!

byrney:
It seems I had a bit more work to do to support all the python versions required. I've marked this as wip while I sort those out.  I'll change the title and drop a comment here when it's passing.

byrney:
I think these are good now. You're right the travis tests don't seem to be 100% reliable. I had 2.7 failing and 3.6 passing. I changed a commit message and now 2.7 passed and 3.6 fails ( but works locally). I've tested 2.6, 2.7, 3.3 and 3.6 locally so fairly confident the travis failures are not real ones.

byrney:
I just noticed that sh is still installed by setup.py. It will be skipped on windows so it's not doing any harm but it's no longer required to run gitlint (only for dev) so it could be removed.  Let me know if you prefer to remove it.

wlgrd:
Any thoughts about moving forward with this?

jorisroovers:
I just merged a branch to master that provides basic support for Windows. The relevant commits were heavily inspired by this PR, so really appreciate the work that was put in here. Thanks again!

In the end, I had to rewrite things slightly to align with some changes that had happened in the meantime and to incorporate some other tweaks.

In any case, I'll be pushing a new release with experimental windows support later in the week. Please use #20 or [any of the other relevant issues](https://github.com/jorisroovers/gitlint/issues?q=is%3Aissue+is%3Aopen+label%3Awindows) for follow-up, I'm closing out this PR.

Issue #67, opened 2018-06-08
work with pre-commit


xsteadfastx:
i really like the gitlint program and i would love to use it in my pre-commit chain. i configured it like in the docs. i just dont get it how it works. how can it check the commit message if its tested pre-commit? do i need to add the commit message to a txt file?

jumanjiman:
@xsteadfastx 

Here's how i use gitlint with pre-commit:
* https://github.com/jumanjihouse/autostager/commit/f50e61533ceb7d73abe00c2049a4e89f67baf38b
* https://github.com/jumanjihouse/docker-rsyslog/commit/3679763aebbd3b99f5669a8996fc3946051f2e45

This allows it to work both as the pre-commit hook and as part of CI.

xsteadfastx:
@jumanjiman does i understand it right that the magic is the stages: commit-msg does the magic?

xsteadfastx:
that does exactly nothing in my case... tried it with `git commit`... my vim opens, i add my message and it commits without gitlint checking it.

in my config:

```
repos:
  - repo: https://github.com/jorisroovers/gitlint
    rev: v0.10.0
    hooks:
      - id: gitlint
        stages: [commit-msg]
  - repo: https://github.com/ambv/black
    rev: stable
    hooks:
      - id: black
        args: [--line-length=88, --safe]
        python_version: python3.6

```

asottile:
@xsteadfastx I think you're missing `pre-commit install --hook-type commit-msg` so that `gitlint` will be applied at message time

PR #72, opened 2018-07-19
Revert "Removed erroneous 'stages' line from .pre-commit-hooks.yaml (#64)"

asottile:
This reverts commit f3906027abb1fd434e9c92b1c7050bf146b3121a.

See https://github.com/jorisroovers/gitlint/pull/64#issuecomment-381410128

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,4 +1,5 @@
 -   id: gitlint
     name: gitlint
-    entry: gitlint --msg-filename
     language: python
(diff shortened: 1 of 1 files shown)


ssbarnea:
I guess this change should fix  https://github.com/jorisroovers/gitlint/issues/67 right?

Apparently in its current form gitlint cannot be used as a pre-commit hook.

asottile:
This is just a revert of a misunderstanding in #64 -- I didn't notice #67 but I'll comment what I think was missing there :)

PR #80, opened 2018-10-10
Integration tests: Include microseconds in temporary git repo path

pbregener:


pbregener:
This was simply a test to check if this one commit (a subset of #76) is enough to fix the intermittent test failures. Not quite, apparently.

Issue #56, opened 2018-02-21
Force using repo via command line option


pbregener:
As discussed in #42 and #52, there seem to be problems with the automatic check for data passed in via STDIN and the fallback to using the local repo. As it appears hard to fix, and some of the proposed (and implemented) fixes only fix the problem for certain CI configurations, I would suggest to add a command line option `--use-repo` (and/or a config file flag) that forces the use of the local repo and ignores any data from STDIN. This would  make use in any CI environment and configuration (however proven or obscure that particular one might be) much easier and fool-proof.

This seems like a boring and simple solution to an apparent otherwise difficult-to-solve problem (see above linked issue).

What do you think @jorisroovers?

jorisroovers:
Yeah, I agree, but would suggest to call it ```--force-local-repo```. Happy to accept .

FWIW, looking at #42  and #52  is a top of mind when I find some proper time to work on gitlint (maybe this weekend?).

pbregener:
As I just wrote in https://github.com/jorisroovers/gitlint/issues/42#issuecomment-377246266, I'm using a pretty ugly but effective workaround right now. I might try to come up with a PR for this, but on the other hand this will be unnecessary if we can get the STDIN check to work properly.

jorisroovers:
I just saw that, pretty clever workaround :-)

I just started looking into #42 and #52 again and am able to reproduce in Jenkins. The fix from #52  doesn't seem to work, but need to spend more time on it - probably later today or tomorrow.

pbregener:
Will be happy to try other approaches in GitLab CI. Just ping me if you have something that's ready for testing :)

PR #87, opened 2019-06-10
Improve handling of piped data in CI environment

rogalski:
Basically, gitlint failed to work successfully on internally-hosted QuickBuild instance.

Apparently "\n" string was piped to gitlint process, which was enough to early exit from "local repository" case. This patch attempts to avoid such case by rejecting whitespace-only piped content. Also, additional debug log is added to improve available debug info when -d switch is used.

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -122,7 +122,7 @@ def get_stdin_data():
         # i.e. don't consider empty piped data
         if input_data:
             stdin_data = ustr(input_data)
-            LOG.debug("stdin data:\n%s", stdin_data)
+            LOG.debug("stdin data:\n%r", stdin_data)
             return stdin_data
     return False
 
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -121,7 +121,9 @@ def get_stdin_data():
         # Only return the input data if there's actually something passed
         # i.e. don't consider empty piped data
         if input_data:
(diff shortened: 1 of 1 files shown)


jorisroovers:
Hi, thanks for your interest in gitlint and creating this PR!

So this is an interesting case, because it gets down into the semantics of what constitutes a git commit message.
While the following example might seem a bit out there, it's technically not impossible to have an white-space only commit message:

```sh
$ git commit --allow-empty-message -m "   " --cleanup=verbatim
$ echo "foo$(git log -1 --pretty=%B)bar"
foo   bar
$ # In this case, gitlint still acts as expected
$  git log -1 --pretty=%B | gitlint
1: T2 Title has trailing whitespace: "   "
1: T6 Title has leading whitespace: "   "
3: B6 Body message is missing
```

I would agree that most people would probably never do this, but then again there are a lot of different git and CI/CD setups (and sometimes it's not humans but machines (not) writing commit messages). The challenge with merging this PR is that there is then no way to still support this whitespace commit example.

I think we might need to look at implementing #56, which is basically a force flag for using the local repo.
Wrt logging the input data for debugging: yes, that should definitely be added :-)

jorisroovers:
We ended up implementing the `--ignore-stdin` flag since v0.12.0 which can be used to deal with this scenario - see #89 and #56. Closing this out.

PR #101, opened 2019-08-15
Tweaked example configuration files

paranoidi:
I think these changes would have made it much more easy for me. What do you think?

I also changed default lengths to follow 50/72 rule, hope you don't mind ;)


jorisroovers:
Thanks - this definitely makes it clearer!

PR #104, opened 2019-09-26
replace readfp with Python2 read_file function, to fix a test failure:

LocutusOfBorg:
```
======================================================================
FAIL: test_config_file (test_cli.CLITests)
Test for --config option
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.7/unittest/mock.py", line 1248, in patched
    return func(*args, **keywargs)
  File "/<<PKGBUILDDIR>>/gitlint/tests/test_cli.py", line 372, in test_config_file
    self.assertEqual(result.output, "")
AssertionError: "/<<PKGBUILDDIR>>/git[197 chars]od\n" != ''
- /<<PKGBUILDDIR>>/gitlint/config.py:366: DeprecationWarning: This method will be removed in future versions.  Use 'parser.read_file()' instead.
-   parser.readfp(config_file, filename)  # pylint: disable=deprecated-method
```

### DIFF
--- a/gitlint/config.py
+++ b/gitlint/config.py
@@ -374,8 +374,7 @@ class LintConfigBuilder(object):
             parser = ConfigParser()
 
             with io.open(filename, encoding=DEFAULT_ENCODING) as config_file:
-                # readfp() is deprecated in python 3.2+, but compatible with 2.7
-                parser.readfp(config_file, filename)  # pylint: disable=deprecated-method
+                parser.read_file(config_file)
 
             for section_name in parser.sections():
                 for option_name, option_value in parser.items(section_name):


LocutusOfBorg:
I don't think Python2 is worth being supported anymore...

LocutusOfBorg:
I'm uploading this patch in Debian, since the package is using Python3 only, feel free to tweak in a retro-compatible way if you think (or fix the test, I don't know which one is better!)

jorisroovers:
Thanks for PR and sorry for late reply...

I think I'm going to support python 2.7 at least throughout 2020. Probably drop it after that.

I'll keep this PR around as reminder to fix this once we drop support for python 2.7.

daniel-baumann:
Hi,

keeping 2.7 compat is fair enough, however, I'd really like to minimize the diff we ship in Debian. How about the following..

Regards,
Daniel
```

diff -Naurp gitlint.orig/gitlint/config.py gitlint/gitlint/config.py
--- gitlint.orig/gitlint/config.py
+++ gitlint/gitlint/config.py
@@ -428,8 +428,11 @@ class LintConfigBuilder(object):
             parser = ConfigParser()
 
             with io.open(filename, encoding=DEFAULT_ENCODING) as config_file:
-                # readfp() is deprecated in python 3.2+, but compatible with 2.7
-                parser.readfp(config_file, filename)  # pylint: disable=deprecated-method
+                if sys.version_info[0] == 2:
+                    # readfp() is deprecated in python 3.2+, but compatible with 2.7
+                    parser.readfp(config_file, filename)  # pylint: disable=deprecated-method
+                else:
+                    parser.read_file(config_file)
 
             for section_name in parser.sections():
                 for option_name, option_value in parser.items(section_name):
```

sigmavirus24:
Why not just use [`parser.read([config_file])`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read) which is present on 2 and 3?

LocutusOfBorg:
I don't think this is needed anymore, better kick it :) feel free to open a new issue if it is still needed

PR #106, opened 2019-10-24
Fix relative links in contributor guide

smoothreggae:
I tested my changes locally using `mkdocs serve`


jorisroovers:
Merged - thanks!

I'll push out the new docs as part of the next release. Not sure when that will be, probably not very not soon, but should be this year.

Issue #130, opened 2020-05-20
New Rule: body-match-regex


strk:
There's a `title-match-regex` rule but no `body-match-regex`, how do I match the whole commit log to make sure a ticket reference exists, no matter if in title or body ?

jorisroovers:
Hi!

That rule indeed doesn't exist. Happy to accept PR for `body-match-regex` and/or `commit-match-regex` :-)

Meanwhile, you should be able to work around this pretty easily using
https://jorisroovers.com/gitlint/user_defined_rules/

strk:
See #131 for `body-match-regex`

PR #131, opened 2020-05-20
Add body-match-regex rule (B8)

strk:
See #130


strk:
Tests are a bit too convoluted for me to write, can you help with that part ?

jorisroovers:
I eventually ended up implementing this feature from scratch in #146 as this PR contained multiple issues and also needed docs and tests as you pointed out.

Thanks for spending the time on creating this PR though! This feature will be released as part of 0.14.0, probably in a few weeks.

PR #133, opened 2020-05-22
Unpin requirements

ssbarnea:
Pinning requirements for no specific reasons creates installation conflicts and also create extra maintenance requests.

As a linter it is expected to be installed with other similar tools.

Closes: #69

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 setuptools
-wheel==0.33.4
-Click==7.0
-sh==1.12.14; sys_platform != 'win32' # sh is not supported on windows
-arrow==0.15.5;
+wheel>=0.33.4
+Click>=7.0
+sh>=1.12.14; sys_platform != 'win32' # sh is not supported on windows
+arrow>=0.15.5;


jorisroovers:
This has come up before, see #69 .
The main reason I don't want to do this is because I've been burned before with dependencies being updated and breaking an otherwise functional gitlint release.

One minute `pip install gitlint` _just works_, the next one it doesn't because some dependency was updated that breaks gitlint. 
Yes, you can work around this by then installing the right dependency versions manually, but truth is that most users don't do this - it's also a unnecessary burden on them IMO.

Then next thing you know, 3 tickets are opened indicating 'gitlint is broken' and the maintainer (i.e. me) has to drop everything to get a fix in. All users then also need to update their scripts, CI pipelines, etc to install the newer version of gitlint. That's a lot of stress and time spent for something that is easily avoided by pinning dependency versions.

Tools like NPM, Pipenv and others also do this by using dependency lock files. This also relates somewhat to https://reproducible-builds.org/, especially the part of deterministic builds.

I know there's strong opposing opinions on this topic, which I believe are based on different experiences (package maintainer vs. users trying to marry conflicting dependency versions). I'm definitely willing to have my opinion changed on this, especially in light of #134, as long as it doesn't mean a lot of extra work or fire-fighting for me :-) 

Interested to hear @sigmavirus24's take on this since he's been following gitlint for a while and gave this PR a thumbs down.

ssbarnea:
Yep, that is one of the dirty topics, being on both sides. In the end it is a gambling whichever path you take. One can imaging that pinning anything would give him security but is not really the case, with recent changes is almost a guarantee that the pinned versions will either crash under newer python versions which did not exist at the time they where pinned or they will be affected by bugs discovered after and already fixed. Changes to packaging ecosystem (pep-517)  are likely to also introduce problems, as you do not have much control over which versions of setuptools/pip/ease_install/virtualenv are used to install the package.

IMHO, the only reproducible way is to package the application as a container. At least that will assure it works the way it worked on the day of the release. But for pypa.org, I am inclined to believe that the "library" path, with controlled compatibility ranges (usually semantic) is bit better.

I happen to know well that `sh` made [important bugfixes ](https://github.com/amoffat/sh/pull/491)in order not to want to use outdated versions. In fact, `sh` maintenance status itself made me strongly consider dropping it on two projects.

I will make another change that is only bumping the pin, as that one can easily be tested and released.

sigmavirus24:
> Interested to hear @sigmavirus24's take on this since he's been following gitlint for a while and gave this PR a thumbs down.

I'm thumbs down specifically because I know how little time you have to spend on this @jorisroovers . Even if you take the Flake8 approach, downstream package managers will still harass and vilify you on social media and file bugs in spite of your issue template asking them not to.

I'm happy to be the big bad upstream developer who would rather protect my time and sanity over having something packaged in a distribution when those tend to move too slowly for developers anyway (who are the target audience for gitlint and flake8, unlike some libraries which use compatibility ranges for similar reasons).

To Sorin's point, pip doesn't actually guarantee that what you specify is what's installed. And personally, given the liberties that some downstream packagers take, I'm not sure that the dependencies listed are anything other than "I as maintainer attest that I've tested against these versions and know, to the best of my knowledge, that they work".

If you go with minimum versions, then your CI has to grow in complexity to test everything and if you pin, then you're reviled. If you use a range, than the next version that [fixes](https://xkcd.com/1172/) something for one user becomes an issue where people pile on complaining that you haven't released in a month.

tl;dr, there's no winning so go with what you're most comfortable with.

neiljp:
I just ran into this while testing the latest `black` in the same environment as `gitlint`, the former seeming to rely upon click>=7.1.2.

I see this PR has `>=`, but have you considered using `~=`? Eg. click~=7.0.0. Unfortunately from what I remember reading in the past this doesn't work properly with `~=7.0`, so wouldn't resolve allowing click 7.x (if that's desirable) but only 7.0 and requiring a new release for 7.1 - but I'd love to hear otherwise!

andersk:
@neiljp `click ~= 7.0.0` and `click ~= 7.0` both work fine, but have different meanings ([documentation](https://www.python.org/dev/peps/pep-0440/#compatible-release)):

* `click ~= 7.0.0` is short for `click >= 7.0.0, == 7.0.*`, so it allows 7.0.2 but not 7.1.2.
* `click ~= 7.0` is short for `click >= 7.0, == 7.*`, so it allows 7.0.2 and 7.1.2 but not 8.1.2.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 setuptools
-wheel==0.33.4
-Click==7.0
-sh==1.12.14; sys_platform != 'win32' # sh is not supported on windows
-arrow==0.15.5;
+wheel>=0.33.4
+Click>=7.0
+sh>=1.12.14; sys_platform != 'win32' # sh is not supported on windows
+arrow>=0.15.5;
```
andersk, line 5: `install_requires` and `extras_require` in `setup.py` also need to be unpinned.

andersk:
> Tools like NPM, Pipenv and others also do this by using dependency lock files. This also relates somewhat to https://reproducible-builds.org/, especially the part of deterministic builds.

The difference is that NPM and Pipenv lock files describe the *entire* package set in use. They don’t cause future conflicts when combined with other requirements because they aren’t combined with other requirements.

> To Sorin's point, pip doesn't actually guarantee that what you specify is what's installed.

This will be changing; `pip install black==20.8b1 gitlint==0.13.1` with Pip 20.2 now prints this warning:

```python3
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

gitlint 0.13.1 requires Click==7.0, but you'll have click 7.1.2 which is incompatible.
Successfully installed appdirs-1.4.4 arrow-0.15.5 black-20.8b1 click-7.1.2 gitlint-0.13.1 mypy-extensions-0.4.3 pathspec-0.8.0 python-dateutil-2.8.1 regex-2020.7.14 sh-1.12.14 six-1.15.0 toml-0.10.1 typed-ast-1.4.1 typing-extensions-3.7.4.3
```

And with `--use-feature=2020-resolver`, which will soon be the default, it fails.

```
ERROR: Cannot install black 20.8b1 and gitlint 0.13.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    black 20.8b1 depends on click>=7.1.2
    gitlint 0.13.1 depends on Click==7.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

wakemaster39:
To add to this, if you are using `poetry` instead of raw pip, gitlint and the latest black are incompatible because of `click==7.0`. 

What adds to the problem is that there is a change in blacks formatting it seems in somewhere in the 20 series compared to the 19, and it is exploding the ability to update pre-commit if I have to force black on a version that click 7.0 is acceptable. 

I understand the pain, but I would do something like `click>=7,click<8` to force major revisions requiring intervention. Its not perfect, and I know dep management is always a disaster, but I hope people one day listen to semver so that it is slightly more sane.

jorisroovers:
This morning I woke up, saw this latest message and thought: *"I understand people are frustrated with gitlint not working together with black, a much more popular tool, let me just do this today and see if I can quickly get 0.13.1 out of the door, even though I really don't have time for it*".

So I tried bumping `click`, `arrow` and `sh` to their latest versions, all minor version bumps.

Turns out, all 3 of those library "minor" version bumps introduce changes that either change gitlint's behavior (arrow) or break gitlint's testing (click, sh). See 7dc9d2b80d9142b19abcc60cfcae2dc368f44314

We can discuss the changes in particular and whether e.g.: gitlint tests should have such strong dependency on the exact output of click, but that's really besides the point. My main point is that I've just spent 2 hours on this, and I still need to resolve an intrusive warning from `arrow` announcing incompatibility with python version < 3.6 and some other failing integration test. That's time I didn't want to spend today, and more importantly that's time I don't want to spend in the future every time a minor version bump breaks/changes either gitlint itself or its testsuite.

Again, I 100% understand the frustation from user's point-of-view, but like @sigmavirus24 eloquently put it: *"there's no winning so go with what you're most comfortable with."* The last few hours proved again that for me this means fixing dependency version numbers fairly tightly.

That said, I'm going to try to get 0.13.1 out of the door over the next week to do the bump to `click~=7.1.2` so the incompatibility with black is resolved. Based on today's challenges, I'm inclined to just bump to a specific minor version and not allow for a version range (maybe allow patch releases).

Appreciate your understanding!

sigmavirus24:
> This morning I woke up, saw this latest message

Let me also put something out there. Every time you see a message from someone saying "I would do X" then let them contribute that if that's really what they want. Let them figure out how to massage everything into it working. You don't owe anyone on this thread anything and I'd argue the overall tone here started with hostility (accusing you of pinning dependencies for no reason is a severe show of bad faith).

> gitlint tests should have such strong dependency on the exact output of click

I'd argue that's not a bad thing to have. So many people script things like this together and parse the output that it changing is a big change for gitlint users who have some kind of automation around that output. It's not unreasonable to want to have some kind of backwards compatibility with that for them, or be able to detect it and release a major version caused by dependency maintainers breaking things in a minor version like appears to have happened here.

tbekolay:
Hi! Sorry to drop into an existing issue and discussion, but we are using `gitlint` and `black` as part of our company's static checks, so unpinning the requirements will soon become necessary for us.

One thing that we have done lately that has lowered the maintenance burden for us significantly is to drop support for old (unsupported) versions of Python. Specifically, requiring Python >= 3.6 has made maintaining our projects more convenient. In the case of this PR, dropping old Python support means that you can upgrade the Click requirement without any issues. Click, I believe, no longer supports 2.7 and 3.5 (see [here](https://github.com/pallets/click/commit/9374365fdd5af45fc3dc4047963fa30aad9ce15c)), so the Python 2.7 regression is not an issue that they are likely to resolve.

There might be some pushback from people who are stuck using old versions of Python. However, I would argue that:

1. There are older releases of `gitlint` that work with those old versions, so if you have to use an old Python version, then sorry but you also have to use an old `gitlint` version.
2. Most of the people using `gitlint` are developers. If they're Python developers, they are likely using a newer version of Python, or can set up a new Python environment quickly if need be. If they're not Python developers, they're likely to be using `gitlint` through a package manager, which will do the hard work of installing a compatible Python version.

I definitely appreciate that your time is limited @jorisroovers; I am in the same boat or else I would readily volunteer to help maintain the project. One suggestion I have is that you could [ask to move the project to the Python Code Quality Authority](https://meta.pycqa.org/en/latest/introduction.html#how-can-i-add-my-project-to-the-pycqa), which would give it some additional visibility, and it would be a bit easier for me to justify helping maintain gitlint to my company.

sigmavirus24:
> One suggestion I have is that you could [ask to move the project to the Python Code Quality Authority](https://meta.pycqa.org/en/latest/introduction.html#how-can-i-add-my-project-to-the-pycqa), which would give it some additional visibility, and it would be a bit easier for me to justify helping maintain gitlint to my company.

So I'm happy to facilitate that move. The issue is that the organization tends to focus on tools that improve the code quality of python projects. Not necessary linters written in Python. There's on-going discussion of what criteria (if any) should be enforced for the authority, and I suspect that if I let everything in, your company wouldn't actually let you work on any of the projects. Even so, I would argue that your company (if they read up on the PyCQA) would realize they probably don't want you working on those projects anyhow as there's no real authority in it.

jorisroovers:
First of all, thanks everyone for your interest and participation - it makes me smile that my small hobby project is getting this amount of attention, especially from people with much more experience in OSS like yourselves.

Second, I think we're starting to have a few discussions at the same time here, can I suggest we split them out?

1. **Unpinning requirements.txt**: Let's keep this issue for that. Right now I'm thinking to not bump them for the upcoming 0.14.x release as that will introduce a few (minor) issues, especially with older python versions. We can bump soon after that, see next item. Unpinning still under debate :)
2. **Python version support**: I had already planned for the upcoming 0.14.0 release to be the last one to support 2.7 and 3.5. Created ticket for related work/discussion: #151
3. **Moving gitlint under a project umbrella**: @ssbarnea had suggested something similar in #134, but moving to @pycontribs. Can we continue this discussion there?

jorisroovers:
Ok, I didn't mean to close this PR, but seems like deleting the master branch (I moved to using `main` as default branch) made this happen automatically. Can't manually reopen this PR either it seems :(

Issue #138, opened 2020-07-21
Add title-min-length rule


tjanez:
It would be similar to the `body-min-length` rule and would quickly detect bad commit messages like "Fix", "Bump", ...

jorisroovers:
Makes sense. Any thoughts on what the best default limitation should be?
This can be done purely by (configurable) char count -  e.g. min 5 chars -  or by needing at least 2 words.

I'm inclined to keep it simple to start and going for min 5 chars.

tjanez:
> Any thoughts on what the best default limitation should be?
> This can be done purely by (configurable) char count - e.g. min 5 chars - or by needing at least 2 words.

Hmm... I also though at doing char count first. At least 2 words probably also makes sense since I can't come up with a good one word commit message example :slightly_smiling_face:.

> I'm inclined to keep it simple to start and going for min 5 chars.

Yes, I think this would be a good start.

PR #140, opened 2020-08-08
New Rule: title-min-length

mrshu:
* Add a new `title-min-length` rule which mimics the `body-min-length`
  rule in its function. The default min length is set to 5 as per the
  discussion in #138.

* Add a few tests modeled on TitleMaxLength.

Signed-off-by: mr.Shu <mr@shu.io>


mrshu:
@jorisroovers thanks a ton for a very nice tool!

I took a quick try on what could the implementation of #138 look like. Please feel free to let me know if there are things that would be worth updating -- should this end up really being part of `gitlint`, it would be nice to mention it in `docs/rules.md` as well for instance.

Thanks again!

cc @tjanez

gitlint/tests/rules/test_title_rules.py
```diff
@@ -152,3 +152,25 @@ def test_regex_matches(self):
...
         self.assertListEqual(violations, [expected_violation])
+
+    def test_min_line_length(self):
+        rule = TitleMinLength()
+
+        # assert no error
+        violation = rule.validate(u"å" * 72, None)
+        self.assertIsNone(violation)
+
+        # assert error on line length > 72
```
jorisroovers, line 163:
```suggestion
        # assert error on line length < 5
```

gitlint/rules.py
```diff
@@ -236,6 +236,20 @@ def validate(self, title, _commit):
...
+class TitleMinLength(LineRule):
+    name = "title-min-length"
+    id = "T8"
+    target = CommitMessageTitle
+    options_spec = [IntOption('min-length', 5, "Min line length in characters")]
+
+    def validate(self, title, _commit):
+        min_length = self.options['min-length'].value
+        actual_length = len(title)
+        if 0 < actual_length < min_length:
```
jorisroovers, line 248: This would still allow for empty commit titles. I'd expect we'd want to capture this too?
```suggestion
        if actual_length < min_length:
```

I think it would be good to add a test for an empty title specifically to cover this scenario

jorisroovers:
This is awesome, thanks so much for taking the time!

Left some comments in review.

I'm currently in a bit of a gitlint hiatus - which will affect when this is released. I *will* do it before the end of 2020, but might be a while... Appreciate your understanding!

mrshu:
Thanks for your response @jorisroovers !

I believe I've applied all the changes you've suggested with two notable exceptions:
1. In the docs I was not sure what `gitlint` version to mention. This can be easily fixed once it comes to the release.
2. I did not include `TitleMinLength` among the default rules for now, as there were quite a few fixtures that would need to get updated to take it into the account. If you think it's still worth doing, I'll update the fixtures as well. 

> I'm currently in a bit of a gitlint hiatus - which will affect when this is released. I will do it before the end of 2020, but might be a while... Appreciate your understanding!

Feel free to take as much time as would suite you -- I really do not think there is any rush here :)

Thanks again!

jorisroovers:
Rebased this to the new `main` default branch. Will be fixing the conflicts and adding the missing code soonish (this weekend hopefully).

mrshu:
Thanks @jorisroovers -- please let me know if I can be of any help in here.

jorisroovers:
Merged! Will go out with 0.14.0, which should be soon!

PR #146, opened 2020-09-11
New Rule: body-match-regex

### LINKED ISSUE #130: New Rule: body-match-regex

jorisroovers:
The body-match-regex rule allows users to enforce that the commit-msg
body matches a given regex.

Closes #130

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -161,6 +161,31 @@ Name                  | gitlint version | Default      | Description
 files                 | >= 0.4          | (empty)      |  Comma-separated list of files that need to an explicit mention in the commit message in case they are changed.
 
 
+## B8: body-match-regex   ##
+
+ID    | Name                        | gitlint version | Description
+------|-----------------------------|-----------------|-------------------------------------------
+B8    | body-match-regex            | >= 0.14         | Body must match a given regex
+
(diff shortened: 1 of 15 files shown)


docs/rules.md
```diff
@@ -161,6 +161,31 @@ Name                  | gitlint version | Default      | Description
...
+#### .gitlint
+
+```ini
+# Ensure the body ends with Reviewed-By: <some value>
+[body-match-regex]
+regex=Reviewed-By:(.*)$
+
+# Ensure body contains the word "Foo" somewhere
+[body-match-regex]
+regex=(*.)Foo(.*)
```
keestux, line 187: The example has two problems.

1. the prefix should be `(.*)`
2. the prefix and postfix are totally unnecessary. The Python code is doing a `search` so the `(.*)` isn't going to make a difference

I suppose it would make more sense to match the word `Foo` (and not `Foolish`) as follows:
```ini
# Ensure body contains the word "Foo" somewhere in the body
[body-match-regex]
regex=\bFoo\b
```

Issue #148, opened 2020-09-23
Error when `user.name` or `user.email` are not set


glasserc:
Hi, `gitlint` is great, I'm trying to convert a project over to use it now (replacing a handful of shell one-liners).

Some checks in the project that I'm trying to convert are to verify that the developer has a name and email set in git. I am not 100% convinced that these checks are valuable but to test them I have been commenting out these variables in my git config file. When I comment out either or both and run gitlint, I get:

`An error occurred while executing '/usr/bin/git config --get user.name': b''`

I guess `gitlint` assumes that name and email are set. For my purposes, that's fine because I'm replacing checks that also ensure that these variables are set. However, I think the error message could be a little clearer.

I'm running `gitlint` through `pre-commit` using a command line like `~/.local/bin/pre-commit run --hook-stage commit-msg --commit-msg-filename test.msg`.

```
/home/ethan/.cache/pre-commit/repok43dfskl/py_env-python3/bin/gitlint --debug --msg-filename test.msg
DEBUG: gitlint.cli To report issues, please visit https://github.com/jorisroovers/gitlint/issues
DEBUG: gitlint.cli Platform: Linux-5.4.0-42-generic-x86_64-with-glibc2.29
DEBUG: gitlint.cli Python version: 3.8.2 (default, Jul 16 2020, 14:00:26) 
[GCC 9.3.0]
DEBUG: gitlint.git ('--version',)
DEBUG: gitlint.cli Git version: git version 2.25.1
DEBUG: gitlint.cli Gitlint version: 0.14.0dev
DEBUG: gitlint.cli GITLINT_USE_SH_LIB: [NOT SET]
DEBUG: gitlint.cli DEFAULT_ENCODING: UTF-8
DEBUG: gitlint.cli Configuration
config-path: /home/ethan/[projectname]/.gitlint
[GENERAL]
extra-path: None
contrib: []
ignore: title-max-length,title-must-not-contain-words,body-max-line-length,body-min-length,body-is-missing
ignore-merge-commits: True
ignore-fixup-commits: True
ignore-squash-commits: True
ignore-revert-commits: True
ignore-stdin: False
staged: True
verbosity: 3
debug: True
target: /home/ethan/Jobs/Teachable/fedora
[RULES]
  I1: ignore-by-title
     ignore=all
     regex=None
  I2: ignore-by-body
     ignore=all
     regex=None
  I3: ignore-body-lines
     regex=None
  T1: title-max-length
     line-length=72
[...]

PR #149, opened 2020-09-24
Fail nicer when username missing

glasserc:
An attempt to make #148 a little less severe. `git` failing to retrieve these configuration variables is still fatal, but at least the error message is a little nicer.


jorisroovers:
Super thanks for creating this PR, I know that must have taken a few hours.

Issue #171, opened 2020-12-20
In a non-interactive environment, git commit doesn't work


PW999:
In a non-interactive git environment (like Visual Studio Code), the git hook hangs while waiting for input from the user.

I've installed gitlint v0.15.0 on OpenSuse in WSL and added it to my repo using install-hook with a default config file using generate-config. However, if I want to commit in Visual Studio Code (using the Remote - WSL extension), the commit process hangs if there's something wrong with my commit message. The only way to proceed is by killing the script

The output

```
> gitlint --staged --msg-filename "$1" run-hook
> git config --get-all user.name
> git config --get-all user.email
> git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file -
> git ls-files --stage -- /home/phillip/src/100-days-of-code/.gitlint
> git cat-file -s 317c4136f08cafaeb8f577a043aa4291684e6424
> git show --textconv :.gitlint
gitlint: checking commit message...
-----------------------------------------------
gitlint: Your commit message contains the above violations.
Continue with commit anyways (this keeps the current commit message)? [y(es)/n(no)/e(dit)] .git/hooks/commit-msg: line 22:  5628 Terminated 
```

I hoped that adding --silent and --ignore-stdin to the commit hook would help but unfortunately it did not.

jorisroovers:
Hi Phillip, nice to see you here :-)

Interesting case, since I have tested this on VScode on Mac when rewriting the commit-msg hook recently (ced7bdec8b941785e7815f2bcaee6b1121efd33d and #157).

I think this might be related to WSL + Remote extension - I've never used those. My gut reaction is the root-cause is probably related to either:

1. The [vodoo we do to check whether we have a TTY connected to stdin](https://github.com/jorisroovers/gitlint/blob/018f42e31c2f8c62771904ee20cfdab088971621/gitlint/files/commit-msg#L4-L8) is somehow failing on WSL + VSCode. I think it's less likely this is the cause.
2. The stdin input is not correctly detected, similar to #91, for which the workaround is to use `--ignore-stdin`. You mentioned having already tried this, but I'd like to double check whether gitlint actually did it right. Can you try enabling debug by adding both `--ignore-stdin` as well as  `--debug` to the gitlint invocation in the `.git/hooks/commit-msg` hook (line 22) and then posting output here?

Thanks!

PW999:
Verified that it was running with the correct arguments using `ps ax` (since I have to kill it), this is the complete command
[...]

PR #172, opened 2020-12-21
Automatically refuse commits with ignore-std-in

PW999:
When ignore-stdin is enabled, do not ask for input but automatically
reject commit messages with violations. This can be usefull for CI/CD
integrations but also for editors like VSCode.

### DIFF
--- a/gitlint/cli.py
+++ b/gitlint/cli.py
@@ -368,7 +368,13 @@ def run_hook(ctx):
             click.echo("-----------------------------------------------")
             click.echo("gitlint: " + click.style("Your commit message contains the above violations.", fg='red'))
 
-            value = None
+            if ctx.obj.config.ignore_stdin:
+                # in a cli or non-interactive environment, automatically refuse the commit and fail with 
+                # the appriopriate exit code
+                value = "n"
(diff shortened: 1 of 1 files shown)


PW999:
Should fix #171

PW999:
This fix seems to have nasty side-effect of failing a lot of tests. Locally the tests ran without issues. 
Wondering if the tests are affected because they are also "non-interactive" and are now behaving differently (but not necessarily incorrectly). 
Going to need some input on it before I start digging in all these unfamiliar test tools 😛

jorisroovers:
The test failures are pep8 (style check) failures, actual unit and integration tests passed. See e.g. https://github.com/jorisroovers/gitlint/pull/172/checks?check_run_id=1590621043

Having said that, I don't think this is the workaround we want. There's a difference between not having a TTY attached to stdin and ignoring stdin all-together.  Let me reply in #171 with more context :-)

PW999:
I already suspected something like that 👍 .

jorisroovers:
Doing some cleanup, closing this PR. Discussion can continue in #171

PR #173, opened 2020-12-21
Error message was confusing with silent enabled.

PW999:
If the silent option was enabled, no violations were printed.
The error message then referred to non-existing "above violations".


PW999:
Should have installed the gitlint commit hook prior to pushing this commit 😄

PR #177, opened 2021-02-12
Fix contrib-body-requires-signed-off-by

gsemet:
Just a typo in the documentation.


jorisroovers:
Tests failed because of missing commit body, fixed that while merging. 

Thanks!

Issue #176, opened 2021-02-10
Blank lines after title cause gitlint to ignore missing body (B6)


dan-cohn-sabre:
I've discovered an unusual problem.  We have folks creating commits with only a title, and gitlint isn't complaining. There's nothing in the .gitlint file to disable this check.

What I've found through testing is that a title followed by 2 or more blank lines passes gitlint even though the message contains no body.

Here's an example with debug:

```bash
$ gitlint --debug --msg-filename ~/COMMIT_EDITMSG run-hook
DEBUG: gitlint.cli To report issues, please visit https://github.com/jorisroovers/gitlint/issues
DEBUG: gitlint.cli Platform: Linux-5.4.39-linuxkit-x86_64-with-centos-8.2.2004-Core
DEBUG: gitlint.cli Python version: 3.6.8 (default, Apr 16 2020, 01:36:27) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]
DEBUG: gitlint.git ('--version',)
DEBUG: gitlint.cli Git version: git version 2.27.0
DEBUG: gitlint.cli Gitlint version: 0.14.0
DEBUG: gitlint.cli GITLINT_USE_SH_LIB: [NOT SET]
DEBUG: gitlint.cli DEFAULT_ENCODING: UTF-8
DEBUG: gitlint.cli Configuration
config-path: /home/sabre2/src/sabre2/.gitlint
[GENERAL]
extra-path: None
contrib: ['contrib-title-conventional-commits']
ignore: 
ignore-merge-commits: True
ignore-fixup-commits: True
ignore-squash-commits: True
ignore-revert-commits: True
ignore-stdin: False
staged: False
verbosity: 3
debug: True
target: /home/sabre2/src/sabre2
[RULES]
  I1: ignore-by-title
     ignore=all
     regex=None
  I2: ignore-by-body
     ignore=all
     regex=None
  I3: ignore-body-lines
     regex=None
  T1: title-max-length
     line-length=72
  T2: title-trailing-whitespace
  T6: title-leading-whitespace
  T3: title-trailing-punctuation
  T4: title-hard-tab
  T5: title-must-not-contain-word
     words=WIP
  T7: title-match-regex
     regex=None
  T8: title-min-length
     min-length=5
  B1: body-max-line-length
     line-length=80
  B5: body-min-length
     min-length=20
  B6: body-is-missing
     ignore-merge-commits=True
  B2: body-trailing-whitespace
[...]

PR #179, opened 2021-02-26
Fix missing body rules with multiple new lines

Lorac:
A body could be crafted with multiple new lines and it would bypass
the body missing rule

It fixes: #176


jorisroovers:
Thanks for the fix, looks perfect. Also verified it locally.

The failing test is an issue with PyPY which I'll need to look at, not related to this. 

Merging. Thanks!

PR #181, opened 2021-02-27
Fix pypy pylint(no-member) failure in CI

Lorac:
pylint is not able to find the member statically. It means that the
lookup member is created dynamically in the pypy implementation


jorisroovers:
Just merged #179 and left a comment in there about this issue. Forgot that you had already fixed it - hero!

Thanks!

Issue #180, opened 2021-02-26
Commit msg not re-read after edit (using run-hook)


marqvart:
Setup:
gitlint as a commit-msg hook (installed using the install-hook command)

Repo steps:

1. A commit msg violating the configured linting rules is used (e.g. git commit -m "This fails linting")
2. gitlint will report the problem(s) and ask the (y)es/(n)o/(e)dit question.
3. edit is chosen, and the message is re-written in the editor (vim in my case). New message is know to pass linting.
4. gitlint will report the same problem(s) as in step 2

Looking at the code, I think the problem is that the message file loaded (via --msg-filename) is kept open (reading position is reset to zero). It is not closed and re-opened after the edit.
My guess is that on Linux, the editor will re-write the file (causing it to get a new inode) while gitlint (via Click) keeps a handle to the old file (old inode). Because the file reading position is just set to zero, the contents of the old file will get re-read after the edit is completed.
(On Windows, this will most likely work because of different file system behavior).

The above behavior on Linux makes the current commit-msg hook implementation non-functional as the user will never be able to fix a linting error using the editor option.

I have reverted to the old (v. 0.10.0) hook implementation (which works fine for 0.15.0 also) as a work-around.

Otherwise: Thanks for a nice tool!

Br
Lars Marqvart

Top output from gitlint --debug:
DEBUG: gitlint.cli To report issues, please visit https://github.com/jorisroovers/gitlint/issues
DEBUG: gitlint.cli Platform: Linux-5.4.0-65-generic-x86_64-with-glibc2.29
DEBUG: gitlint.cli Python version: 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0]
DEBUG: gitlint.git ('--version',)
DEBUG: gitlint.cli Git version: git version 2.25.1
DEBUG: gitlint.cli Gitlint version: 0.15.0
DEBUG: gitlint.cli GITLINT_USE_SH_LIB: [NOT SET]
DEBUG: gitlint.cli DEFAULT_ENCODING: UTF-8
DEBUG: gitlint.cli Configuration

jorisroovers:
Thanks for opening the issue - interesting case.

The fix in #182 looks good, but as noted in there: we probably need a test case to capture regressions around this and maybe do some additional manual testing on windows to make sure. 

I should have time in the coming week to get this merged and a fix released :-)

jorisroovers:
[...]

PR #182, opened 2021-02-28
Fix msg-filename kept open after edition

Lorac:
--msg-filename type is now a Path and not a managed click.File.
This makes it easier to control the opening and closing of the file.

Context manager will make sure that the file we use will be refreshed after
edition.

Fixes: #180


Lorac:
@jorisroovers I would like to add a test for this, but I think I'll need a pointer.

jorisroovers:
Thanks for this fix. At first sight, it definitely looks good to me.

As you probably saw, tests failing is because of import-order issue (easy fix):

```bash
gitlint/cli.py:10:0: C0411: standard import "import pathlib" should be placed before "import click" (wrong-import-order)
```

I agree we probably do need a test specific for this scenario, to avoid regressions in the future. 

i.e. a test cases that: 

1. Installs git commit message hook
2. Uses a commit message with violations, checks that violations are as expected
3. Edits the commit message and fixes violations using the hook edit prompt
4. Checks that no violations are reported

Testing the hook end-to-end has always been a tricky though - the related test code might be a bit of a head scratcher initially.

Some pointers:

- Closest unit test, probably needs to be duplicated and modified: [https://github.com/jorisroovers/gitlint/blob/c175e45b8ce14af9e55c442a30c95ecc65fcee68/gitlint/tests/cli/test_cli_hooks.py#L125](https://github.com/jorisroovers/gitlint/blob/c175e45b8ce14af9e55c442a30c95ecc65fcee68/gitlint/tests/cli/test_cli_hooks.py#L125)
- Closest integration/system test. These invoke gitlint from the commandline, so no internal gitlint code is tested, only behavior. Same as unit test: would need to be duplicated and edited. [https://github.com/jorisroovers/gitlint/blob/c175e45b8ce14af9e55c442a30c95ecc65fcee68/qa/test_hooks.py#L110](https://github.com/jorisroovers/gitlint/blob/c175e45b8ce14af9e55c442a30c95ecc65fcee68/qa/test_hooks.py#L110)
- Either a unit or integration tests is fine, I don't think we need both

I'm happy to spend some time myself, please let me know, I should have time in the next few days.

Thanks!

Lorac:
@jorisroovers Thanks for looking at it. I added a test. If you know a way to make it better I would like to know.

jorisroovers:
Superceded by #491, closing this one.

PR #188, opened 2021-05-07
Add missing environment variable for config file

Notgnoshi:
Most of the gitlint command-line options have corresponding environment
variables following the format GITLINT_<commandline_option>, added in
69aa5ffaeb7dae849f8ab655ea0426e26deee44f.

---

The only remaining command-line option that's missing a corresponding environment variable is the `-c` option for overriding individual configuration options.

Additionally, I made the assumption in the documentation that this change would get released in 1.16.0.

I investigated adding tests, but I chose not to for a few, opinionated, reasons
* This is a quite small change, that follows an existing pattern
* The most similar environment variable, `GITLINT_EXTRA_PATH`, does not have a test
* I did not find a clean way to add a meaningful test, but I'm not very experienced with Python testing methods that interact with the system environment.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -318,13 +318,13 @@ Path where gitlint looks for a config file.
 
 Default value              |  gitlint version | commandline flag  | environment variable   
 ---------------------------|------------------|-------------------|-----------------------
- (empty)                   | >= 0.18.0        | `--config`        | `GITLINT_CONFIG`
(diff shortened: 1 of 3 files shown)


jorisroovers:
Made a few minor doc tweaks and merged! Thanks for submitting the PR!

Not sure when I'll push out the next release, likely to be weeks, maybe few months...

PR #195, opened 2021-07-28
docs: Use standard virtualenv setup instructions

l0b0:
A few tweaks to the local development environment setup docs.


jorisroovers:
Thanks!

PR #197, opened 2021-07-28
tests: Set minimum test coverage

l0b0:
This ensures that code which significantly reduces the coverage won't be
accepted.


jorisroovers:
Thanks!

PR #200, opened 2021-07-28
feat: Configure Dependabot

l0b0:
It might still need to be enabled in the repo configuration.


jorisroovers:
Thanks, I should've done this a long time ago!

Issue #196, opened 2021-07-28
test_run_hook_edit is not portable


l0b0:
`test_run_hook_edit` relies on the default `$EDITOR` value of the CI environment, which is not portable. For example, on Arch Linux the test fails with the following error:

```text
E           AssertionError: expected call not found.
E           Expected: shell('vim -n /tmp/tmpj8g6mqhp/hür')
E           Actual: shell('vim /tmp/tmpj8g6mqhp/hür')
```

jorisroovers:
Actually, that test isn't meant to depend on the environment, it overrides `os.environ['EDITOR']` here:

https://github.com/jorisroovers/gitlint/blob/4a451333cf59fdb078ec6859500cea418df434de/gitlint/tests/cli/test_cli_hooks.py#L128-L134

That first `set_editor[0]` set to `None` in particular is meant to test for the default test case where no `$EDITOR` is set and gitlint falls back to `vim -n`, see following 2 snippets.

https://github.com/jorisroovers/gitlint/blob/4a451333cf59fdb078ec6859500cea418df434de/gitlint/cli.py#L28

https://github.com/jorisroovers/gitlint/blob/4a451333cf59fdb078ec6859500cea418df434de/gitlint/cli.py#L396

I'm not sure why this test is failing for you then, there might be something wrong in my testing logic. Would need to dig in a bit deeper.

FYI that I do have it on the [roadmap](https://github.com/jorisroovers/gitlint/wiki/Roadmap) to respect `git config core.editor` instead of having a hard coded `$EDITOR` fallback, just haven't gotten around to that.

l0b0:
Interesting! FWIW, here's an extract of my settings:

- `git config core.editor` is not set
- `$EDITOR` is just `vim`
- Up-to-date version of Arch Linux
- Python 3.8.10

Test results on your current main branch (commit 1a0ed8d, "Bump arrow from 1.0.3 to 1.1.1 (#210)"), of which it looks like many more are not portable:

```text
$ ./run_tests.sh --all
### PYTHON (Python 3.8.10, /home/victor/dev/gitlint/.venv/bin/python) ###
# UNIT TESTS (Python 3.8.10, /home/victor/dev/gitlint/.venv/bin/python) #
Cleaning the *.pyc, site/, build/, dist/ and all __pycache__ directories...DONE
=========================================== test session starts ============================================
platform linux -- Python 3.8.10, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
[...]

PR #201, opened 2021-07-29
feat: Configure shellcheck pre-commit hook

l0b0:
And fix reported shellcheck lints.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -36,17 +36,17 @@ NO_COLOR="\033[0m"
 
 title(){
     MSG="$BLUE$1$NO_COLOR"
-    echo -e $MSG
+    echo -e "$MSG"
 }
 
 subtitle(){
     MSG="$YELLOW$1$NO_COLOR"
-    echo -e $MSG
+    echo -e "$MSG"
 }
 
 fatal(){
     MSG="$RED$1$NO_COLOR"
-    echo -e $MSG
+    echo -e "$MSG"
     exit 1
 }
 
@@ -61,7 +61,7 @@ handle_test_result(){
     EXIT_CODE=$1
     RESULT="$2"
     # Change color to red or green depending on SUCCESS
-    if [ $EXIT_CODE -eq 0 ]; then
+    if [ "$EXIT_CODE" -eq 0 ]; then
         echo -e "${GREEN}SUCCESS"
     else
         echo -e "${RED}FAIL"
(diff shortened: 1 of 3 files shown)


jorisroovers:
Thanks for this PR! 

Had considered this before, just not gotten to it. Also, I'm still debating whether to move to python entirely for task/test running. I had experimented a long time back on a separate branch using [pyinvoke](http://www.pyinvoke.org/), but plain python might be better choice since it doesn't add an extra dependency... See https://github.com/jorisroovers/gitlint/commit/f50390ab187c1cc512037fd5b25295c3db2c59f3

Specific to this PR:
1. Let's not include pre-commit as part of this PR. That's an entirely separate topic IMO. I'd prefer to add a new `./run_tests.sh -ls` flag that invokes shellcheck and then add it a step to our github CI action to invoke that: https://github.com/jorisroovers/gitlint/blob/main/.github/workflows/checks.yml
2. Tests failed on `./run_tests.sh -p`, I think the PR contains a bug. 
3. We need make sure we manually test all `./run_tests.sh` subcommands before we merge this. Some of those are not part of CI runs.

Thanks!

l0b0:
> Had considered this before, just not gotten to it. Also, I'm still debating whether to move to python entirely for task/test running. I had experimented a long time back on a separate branch using [pyinvoke](http://www.pyinvoke.org/), but plain python might be better choice since it doesn't add an extra dependency... See [f50390a](https://github.com/jorisroovers/gitlint/commit/f50390ab187c1cc512037fd5b25295c3db2c59f3)

I can thoroughly recommend testing in Python. For one thing, Python is superior to Bash for basically anything to do with testing (I've used shunit2 a fair bit and Bash quite a lot).
 
> Specific to this PR:
> 
>     1. Let's not include pre-commit as part of this PR. That's an entirely separate topic IMO.

That's a good point; removed. However:

> I'd prefer to add a new `./run_tests.sh -ls` flag that invokes shellcheck and then add it a step to our github CI action to invoke that: https://github.com/jorisroovers/gitlint/blob/main/.github/workflows/checks.yml

I personally don't like this solution. There's a giant shell script in front of all the test logic, when `pre-commit run --all-files` and `pytest` could do all or almost all of it.

>     3. We need make sure we manually test all `./run_tests.sh` subcommands before we merge this. Some of those are not part of CI runs.

Why not? I ran `./run_tests.sh --all`, but [a bunch of those tests fail even on the main branch](https://github.com/jorisroovers/gitlint/issues/196#issuecomment-899186300), so I've no idea whether my changes broke things.

l0b0:
Ping?

l0b0:
Rebased.

jorisroovers:
So I still think I want to move away from bash/shell and move to pure python for checks and tests. Perhaps moving to [hatch](https://github.com/pypa/hatch), [tox](https://tox.wiki/en/latest/) or something else still. I plan to look into this before the end of the calendar year. 

Most or all of `run_tests.sh` would go away as a result. With the limited time I have for gitlint, I don't think investing it in this PR is then the best thing. Appreciate the effort you put in here though, thanks!

l0b0:
No worries. Would you be interested in [other pre-commit hooks](https://github.com/linz/geostore/blob/master/.pre-commit-config.yaml) and/or [automated linting](https://github.com/linz/geostore/blob/c3c2e1a3695be9c3403761ec46a3a542cda837d3/.github/workflows/test.yml#L7-L49)?

jorisroovers:
For me, CI and the move away from `run_tests.sh` kind of go hand in hand and as such I prefer to wait a little bit.  There's definitely a few interesting ones (some I've used before) in what you linked. Today we don't use pre-commit, that might change - allow me some time to consider this :-)

PR #214, opened 2021-08-19
fix: Not fail when current branch has no commits.

### LINKED ISSUE #189: Gitlint requires at least one commit to function
### The Problem

*Gitlint version: 0.15.1*

When linting a staged commit with `--staged` in an empty repo or orphan branch with no commits a `GitContextError` is raised with the message:

> Current branch has no commits. Gitlint requires at least one commit to function.

While the error is clear, I feel that under the circumstances gitlint should give a little leeway, especially since the exception in this case comes down to a debug log line in **gitlint/lint.py** (when removed, gitlint works without issue):

```
 10 class GitLinter:
    ...
 69     def lint(self, commit):
 70         """ Lint the last commit in a given git context by applying all ignore, title, body and commit rules. """
 71         LOG.debug("Linting commit %s", commit.sha or "[SHA UNKNOWN]")
 72         LOG.debug("Commit Object\n" + str(commit))
                                          ^
```
In **gitlint/git.py**:

```
127 class GitCommit:
    ...
162     def __str__(self):                                                                       
163         date_str = arrow.get(self.date).format(GIT_TIMEFORMAT) if self.date else None
164         return (f"--- Commit Message ----\n{self.message}\n"
165                 "--- Meta info ---------\n"
166                 f"Author: {self.author_name} <{self.author_email}>\n"
167                 f"Date:   {date_str}\n"
168                 f"is-merge-commit:  {self.is_merge_commit}\n"
169                 f"is-fixup-commit:  {self.is_fixup_commit}\n"
170                 f"is-squash-commit: {self.is_squash_commit}\n"
171                 f"is-revert-commit: {self.is_revert_commit}\n"
172                 f"Branches: {self.branches}\n"
                                 ^
173                 f"Changed Files: {self.changed_files}\n"
174                 "-----------------------")
```

`StagedLocalGitCommit.branches` attempts to get the current branch name by calling `git rev-parse --abbrev-ref HEAD` which fails.

#### 1) Be forgiving of `GitContextError` in `gitlint.git.StagedLocalGitCommit.branches`

Catch the exception and return an empty list.  I don't know where this property is normally used other than debugging messages but if users can handle an empty list then this seems reasonable.

#### 2) Use `git branch --show-current` to get the branch name

The `--show-current` option was added to git in 2.22.0 (June 2019) so may be a little new to rely on.  It can also return empty output for detached HEAD which will need to be handled.

hongqn:
fixes #189.

It uses `git branch --show-current` to get current branch when the old way (`git rev-parse --abbrev-ref HEAD`) fails.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: "ubuntu-latest"
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", pypy3]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", pypy-3.9]
         os: ["macos-latest", "ubuntu-latest"]
     steps:
       - uses: actions/checkout@v3
@@ -21,7 +21,7 @@ jobs:
         run: mv .git ._git
 
       - name: Setup python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v3.1.1
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,7 +21,7 @@ jobs:
         run: mv .git ._git
 
       - name: Setup python
-        uses: actions/setup-python@v3.1.1
+        uses: actions/setup-python@v3.1.2
         with:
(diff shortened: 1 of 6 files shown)


jorisroovers:
Verified locally, refactored the tests to better match existing tests. Thanks for the PR, merged!

PR #226, opened 2021-10-04
Testing whether gitlint check is skipped on PRs

jorisroovers:
Just testing whether gitlint check is skipped on PRs


jorisroovers:
So all the PR CI jobs are being skipped. The ones that are failing are those that ran on the push to the topic branch I created for this test. This issue won't be there when a PR is coming from a fork. 

We could put in an extra condition to not run the gitlint check on other branches than main, but I think this will suffice for now.

PR #238, opened 2021-10-29
Fallback to $HOME/.gitlint

haoadoresorange:
Maybe not making this default behaviour but adding an option to `install-hook` ?

EDIT: typo


jorisroovers:
I think this can be solved more generically using #188, without the need to change the hook script.

haoadoresorange:
indeed, I'll close this then.

PR #242, opened 2021-11-05
Fixed recogize merge commit fail while merge commit's parents is empty

HuangKaiHuan:
os: ubuntu16.04
git: 2.33.1
gitlint: 0.16.0

when i use gitlint --commits HEAD to check the commit-msg
one merge commit check the format fail, but the others will not.
I had set ignore-merge-commits=true.
I found the parents of this commit is empty.
and then, I use git git log 273b7ecde3a1151185b70244febc40a3f2049e15 -1 --pretty=%P to check,
the parents is still empty.

now, we set:
commit_is_merge_commit = len(commit_parents) > 1
can we set:
commit_is_merge_commit = len(commit_parents) > 1 or commit_msg.startswith("Merge")

the last,
the error occurs, only when gitlint runs on gitlab-ci(docker)
it never occurs, then gitlint runs on my local pc.

self.__class__.__name__: LocalGitCommit {'message': <gitlint.git.GitCommitMessage object at 0x7fb9dd7540d0>, 'author_name': 'huangkh', 'author_email': 'huang_kh@robotics-robotics.com', 'date': datetime.datetime(2021, 9, 1, 13, 12, 25, tzinfo=tzoffset(None, 28800)), 'parents': ['273b7ecde3a1151185b70244febc40a3f2049e15'], 'is_merge_commit': False} {'message': <gitlint.git.GitCommitMessage object at 0x7fb9dd7540d0>, 'author_name': 'huangkh', 'author_email': 'huang_kh@robotics-robotics.com', 'date': datetime.datetime(2021, 9, 1, 13, 12, 25, tzinfo=tzoffset(None, 28800)), 'parents': ['273b7ecde3a1151185b70244febc40a3f2049e15'], 'is_merge_commit': False, 'branches': ['(HEAD detached at 441231c)'], 'changed_files': ['src/rrdevices/three_color_lamp/__init__.py', 'src/rrdevices/three_color_lamp/base_color_lamp.py', 'src/rrdevices/three_color_lamp/ddr_color_lamp.py', 'src/rrdevices/three_color_lamp/kent_color_lamp.py', 'tests/test_color_lamp.py']} try cache: fFalse parents: commit_msg: Merge branch 'io_board' into 'master' feat: add io_board ui See merge request hkh/rrdevices!1 commit_parents: [''] commit_is_merge_commit: False commit_msg_obj: Merge branch 'io_board' into 'master' feat: add io_board ui See merge request hkh/rrdevices!1

### DIFF
--- a/gitlint/git.py
+++ b/gitlint/git.py
@@ -206,7 +206,7 @@ class LocalGitCommit(GitCommit, PropertyCache):
         (name, email, date, parents), commit_msg = raw_commit[0].split('\x00'), "\n".join(raw_commit[1:])
 
         commit_parents = parents.split(" ")
-        commit_is_merge_commit = len(commit_parents) > 1
+        commit_is_merge_commit = len(commit_parents) > 1 or commit_msg.startswith("Merge")
 
         # "YYYY-MM-DD HH:mm:ss Z" -> ISO 8601-like format
         # Use arrow for datetime parsing, because apparently python is quirky around ISO-8601 dates:


sigmavirus24:
I hadn't ever seen a case with a merge commit with only 1 parent in practice before but I knew it was theoretically possible. https://stackoverflow.com/questions/22218548/how-could-this-git-merge-commit-can-have-1-parent-only seems that someone did a rebase without `--preserve-merges`

jorisroovers:
So gitlint already uses this heuristic when you pass a message to it via stdin, like so:

```sh
git log -1 --pretty=%B | gitlint
```

I think this can work as a workaround for you.

When reading from the local git repo, we don't use this heuristic as you pointed out. This is on purpose, to avoid having false positives when someone uses a commit message that starts with "Merge", e.g. "Merge sort implemented".

Additionally, I think you can argue that if a commit only has a single parent, it's not really a merge commit anymore. Clearly the original commit author didn't intend it to be if they went through the effort to hide it. Caveat: I didn't really go into the depths of this much further, but I think the point remains that a commit message starting with "Merge" alone is not enough for it to be a merge commit.

As an alternative workaround, you can write a [user-defined configuration rule](https://jorisroovers.com/gitlint/user_defined_rules/#configuration-rules) that changes this behavior.

HuangKaiHuan:
Thank

PR #220, opened 2021-09-17
Make requirement pinning optional

andersk:
gitlint’s pinned requirements make it difficult to use with other libraries and tools that have newer requirements, and may hold back important security updates. With this change, it’s possible to install gitlint with its requirements either pinned:

```sh
pip install gitlint[pinned]
```

or unpinned:

```sh
pip install gitlint
```

At this time, we document only the pinned option because the maintainer has indicated a preference for supporting that configuration.

Fixes #162.

PR #246, opened 2021-11-13
Split into gitlint and gitlint-core to allow unpinning requirements

### LINKED ISSUE #162: Unpin requirements.txt
This issue can be used as a continuation of #133 which I inadvertently (and irreversibly) closed when deleting the `master` branch in favor of the new `main` default branch.

**Please read ALL of #133 before commenting here.**

andersk:
`gitlint`’s pinned requirements make it difficult to use with other libraries and tools that have newer requirements, and may hold back important security updates. However, the maintainer prefers to expose pinned requirements by default.

Reconcile this by splitting `gitlint` into two packages:

* `gitlint-core` has unpinned requirements by default, but pinned requirements with the `[trusted-deps]` extra.
* `gitlint` becomes an empty package that requires `gitlint-core[trusted-deps]`.

Fixes #162. Alternative to #220.


gitlint-core/setup.py
```diff
@@ -60,12 +60,15 @@ def get_version(package):
     ],
     python_requires=">=3.6",
     install_requires=[
-        'Click==8.0.3',
-        'arrow==1.2.1',
+        'Click>=8',
+        'arrow>=1',
```
sigmavirus24, line 64: This seems unnecessarily wide, why not `>=1.2` Same for `sh>=1.14`

andersk: Gitlint uses a tiny subset of the arrow API that works as far back as 0.3.2.

I’ll go ahead and bound `sh>=1.13.0` since `_unify_ttys` is used in the `qa` tests—even though we still run tests with pinned requirements, and the program itself works fine as far back as sh 1.05, the earliest version supporting Python 3.3+.

sigmavirus24:
Now that there's no top-level `setup.{py,cfg}` does pre-commit still work with the designed `.pre-commit-hooks.yaml` in the repository or does that need additional work to support `pre-commit` users? What does build & release look like now? I suspect it's not as straight forward as before. 

@jorisroovers what's your workflow for releasing gitlint today?

andersk:
Fixed `pre-commit` by moving the `gitlint` package back to top level and adjusting `.pre-commit-hooks.yaml`. (Users will not need to change anything.)

Build and release will look similar to what it looks like now, it will just be done in both directories, e.g.

```shell
python setup.py sdist bdist_wheel upload
cd gitlint-core
python setup.py sdist bdist_wheel upload
```

jorisroovers:
Thanks @andersk for doing all this work :)

I've reviewed this and tried it out locally - as far as I can tell, everything looks good and is working. I didn't test pre-commit, so @sigmavirus24, appreciate if you can confirm that's working after the latest changes. I'll also wait for your overall thumbs up before merging.

Step-by-step release procedure is documented on the wiki: [https://github.com/jorisroovers/gitlint/wiki/Release-MOP](https://github.com/jorisroovers/gitlint/wiki/Release-MOP)

I think @andersk is correct that we only need to change the build and upload steps, I'll get the wiki updated during the next release.

I think we'll want to update the package description of `gitlint-core` and add some more details on this split in our Contributing documentation, but those are incremental improvements we can do in the future.

Once merged, I'll get this released over the next week or so - just a matter of making some time.

jorisroovers (approved):

sigmavirus24 (approved):
I confirmed this works with pre-commit doing:

```yaml
- repo: https://github.com/andersk/gitlint
  rev: core
  hooks:
  - id: gitlint
```

I'm not a fan of using static version strings in _two_ different places, but I think @jorisroovers and I can fix that up after merging this.

jorisroovers:
Thanks @sigmavirus24. Actually had the same thought about that version string, let's indeed fix that afterwards. 

Merged this PR. Thanks @andersk!
I should have time over the next week to release 0.17.0 with these changes.

PR #251, opened 2021-12-01
Add changelog project URL to project metadata

scop:
For easy/"standard" access from the PyPI project page.

`Changelog` matches the title of the linked doc here, and is one of
the recognized names that get the distinctive icon:
https://github.com/pypa/warehouse/blob/64102a6a41f024e54bc0cfe52201d44bc80afd17/warehouse/templates/packaging/detail.html#L24

An example how it looks can be found e.g. in https://pypi.org/project/black/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ setup(
     project_urls={
         'Documentation': 'https://jorisroovers.github.io/gitlint',
         'Source': 'https://github.com/jorisroovers/gitlint',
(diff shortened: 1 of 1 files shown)


jorisroovers:
Great, thanks!

Just realized we should probably do this for gitlint-core as well, in this file:
https://github.com/jorisroovers/gitlint/blob/main/gitlint-core/setup.py

PR #256, opened 2022-01-04
Unset EDITOR env var in test_run_hook_edit

carlsmedstad:
Without this the test will fail if the `EDITOR` env var is set to
something different than `vim -n`.


jorisroovers:
Thanks for the PR!

PR #263, opened 2022-02-06
Update incorrect uses of 'title' in docs/rules.md

jtaylor100:
Small correction of the docs.


sigmavirus24 (approved):
Thanks for catching and fixing these!

PR #295, opened 2022-06-06
Fix descriptions of length config parameters

scop:
They're all inclusive.


jorisroovers:
Great catch! Will merge once CI issues are resolved 👍

PR #313, opened 2022-08-01
Add check to block GitHub author information

matthiasbeyer:
Another contrib rule to block GitHub author information in commits,... some people really use these, can you believe it?

---

I'm not a python guy, you might want to look closely :laughing:

### DIFF
--- a/gitlint-core/gitlint/contrib/rules/block-github-author.py
+++ b/gitlint-core/gitlint/contrib/rules/block-github-author.py
@@ -1,0 +1,32 @@
+import re
+
+from gitlint.options import ListOption
+from gitlint.rules import CommitMessageTitle, LineRule, RuleViolation
+
+RULE_REGEX = re.compile(r"(.*)users\.noreply\.github\.com(.*)")
+
+
+class BlockGithubAuthor(CommitRule):
+    """ This rule enforces that no github author information (email address) is
+    used for committing
+    """
+
+    name = "contrib-block-github-author"
+    id = "BG1"
+
+    def validate(self, line, commit):
+        if commit.is_merge_commit:
+            # We do not care about merge-commits here
+            return []
+
+        if commit.author_email is None:
+            return [RuleViolation(self.id, f"No author information in commit", line)]
+        else:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Doesn't this block merge commits generated by GitHub if someone has opted to make their email private?

matthiasbeyer:
It has a line for ignoring merge commits.

Maybe we can make this exception even configurable. Some projects may want to enforce their contributors to have their information public.

sigmavirus24:
> It has a line for ignoring merge commits.
> 
> Maybe we can make this exception even configurable. Some projects may want to enforce their contributors to have their information public.
> 

They might, but regardless saying "who would even do - this?" (Or whatever you said exactly - I'm on my phone) is incredibly shitty. People have reasons. Don't judge them just because you won't try to look at things from their perspective

matthiasbeyer:
Agreed. That's why I wrote this check in the first place! Because people might have a good reason to not allow anonymous contributions! It goes both ways, you know?

jorisroovers:
This PR has multiple issues (also see #312 for similar feedback)
- Missing tests and docs
- Not following existing rule naming conventions
- Implementing multiple things at once (ignoring merge commits, disallowing empty authors, disallowing a specific email address)
- Regex is not user-configurable (it should probably be a rule option)

Regardless, to me this looks like a variation of https://jorisroovers.com/gitlint/rules/#m1-author-valid-email

I think you can use that rule either in combination with a [negative regex match](https://www.adamsmith.haus/python/answers/how-to-do-negative-pattern-matching-using-regular-expressions--in-python) (I tried for 30 sec and gave up, but I think it might be possible), or alternatively we could implement a rule-option for that rule that enables negative matching.

matthiasbeyer:
> Regardless, to me this looks like a variation of https://jorisroovers.com/gitlint/rules/#m1-author-valid-email

Ah, yes, I missed that one. I will close this PR as I think this is just (as you said) a variation of the linked rule and adapting the existing one should IMO be preferred over duplicating stuff! Thanks for taking the time!

PR #322, opened 2022-08-15
Add check to block GitHub author information

matthiasbeyer:
This adds a new rule for blocking github generated commits from a PR.

Usecases:

* Block commits that were generated from "Suggestions" but not rebase-squashed
* Block github web-ui commits / commits without a user email address.

The latter point might be relevant for projects because of legal issues - if a PR author does not provide contact information, changes from them might be not acceptable for certain projects.

---

The commit(s) from this PR were made before or during my initial efforts in #312, so obviously this is all not up to date regarding details like rule ID and so on.
That's why I'm filing this as a draft for now, so that you can first review the general idea and if you reject this entire idea, I do not waste effort :smile:

### DIFF
--- a/gitlint-core/gitlint/contrib/rules/block-github-author.py
+++ b/gitlint-core/gitlint/contrib/rules/block-github-author.py
@@ -1,0 +1,32 @@
+import re
+
+from gitlint.options import ListOption
+from gitlint.rules import CommitMessageTitle, LineRule, RuleViolation
+
+RULE_REGEX = re.compile(r"(.*)users\.noreply\.github\.com(.*)")
+
+
+class BlockGithubAuthor(CommitRule):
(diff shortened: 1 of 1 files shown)


jorisroovers:
How is this different from #313?

Specifically:

> I think you can use that rule either in combination with a [negative regex match](https://www.adamsmith.haus/python/answers/how-to-do-negative-pattern-matching-using-regular-expressions--in-python) (I tried for 30 sec and gave up, but I think it might be possible), or alternatively we could implement a rule-option for that rule that enables negative matching.

matthiasbeyer:
Meh, I forgot that I already filed it, that there was a solution and that I just did not clean up my branch locally.

Sorry about that :laughing:

PR #335, opened 2022-09-19
fix: Use specific checkout action version

l0b0:
GitHub Actions unfortunately runs the *latest* version of the actions which fits the `uses` directive version. So right now `actions/checkout@v3` is equivalent to `actions/checkout@v3.0.2`, but this can change at any time. To avoid workflows suddenly failing in case a new minor/patch version of an action, this PR uses the most specific version number available for the checkout action.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -10,7 +10,7 @@ jobs:
         python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", pypy-3.9]
         os: ["macos-latest", "ubuntu-latest"]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v3.0.2
         with:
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I don't think we want to pin this strictly. Checkout shouldn't fail randomly and if it does, I'd expect a newer version to be released quickly

jorisroovers:
No super strong opinion on this. 

FWIW, with python we did pin to a specific version and dependabot has been doing a good job of bumping these, for example: https://github.com/jorisroovers/gitlint/pull/316

However small, pinning does reduce the chance for failure because of version upgrades, as such, I don't think it hurts to merge this PR.

jorisroovers:
Again, no strong opinion - I just merged to clean this up. We can easily rollback if it's not working for us.

PR #350, opened 2022-10-25
Python 3.11 release support

jorisroovers:
Python 3.11 was released on 2022-10-24, enable support in CI for
the final 3.11 release (was RC2 before).


jorisroovers:
3.11 not released yet for github CI.
Should pop up here once it does: https://github.com/actions/python-versions/releases

PR #358, opened 2022-11-03
Add authors-commit contrib rule

stauchert:
It is possible to manage a so-called [AUTHORS](https://opensource.google/documentation/reference/releasing/authors) file in the repository. In this file, all committers for the respective project are listed.
With this `gitlint` rule, it can be ensured that only people who are listed in the AUTHORS file can commit.

### DIFF
--- a/gitlint-core/gitlint/contrib/rules/authors_commit.py
+++ b/gitlint-core/gitlint/contrib/rules/authors_commit.py
@@ -1,0 +1,48 @@
+import re
+from pathlib import Path
+from typing import Tuple
+
+
+from gitlint.rules import CommitRule, RuleViolation
+
+
+class Authors(CommitRule):
+    """Enforce that only authors listed in the AUTHORS file are allowed to commit."""
+
+    authors_file_names = ("AUTHORS", "AUTHORS.txt", "AUTHORS.md")
+    parse_authors = re.compile(r"^(?P<name>.*) <(?P<email>.*)>$", re.MULTILINE)
+
+    # A rule MUST have a human friendly name
+    name = "contrib-authors-commit"
+
+    # A rule MUST have a *unique* id, we recommend starting with UC (for User-defined Commit-rule).
+    id = "CC3"
+
+    @classmethod
+    def _read_authors_from_file(cls, git_ctx) -> Tuple[str, str]:
+        for file_name in cls.authors_file_names:
+            path = Path(git_ctx.repository_path / file_name)
+            if path.exists():
+                authors_file = path
+                break
+        else:
+            raise FileNotFoundError("No AUTHORS file found!")
+
+        authors_file_content = authors_file.read_text("utf-8")
+        authors = re.findall(cls.parse_authors, authors_file_content)
+
+        return set(authors), authors_file.name
+
+    def validate(self, commit):
+        registered_authors, authors_file_name = Authors._read_authors_from_file(commit.message.context)
+
+        author = (commit.author_name, commit.author_email.lower())
+
+        if author not in registered_authors:
+            return [
+                RuleViolation(
+                    self.id,
+                    f"Author not in '{authors_file_name}' file: " f'"{commit.author_name} <{commit.author_email}>"',
+                )
+            ]
+        return []
--- a/gitlint-core/gitlint/contrib/rules/authors_commit.py
+++ b/gitlint-core/gitlint/contrib/rules/authors_commit.py
@@ -21,7 +21,7 @@ class Authors(CommitRule):
     @classmethod
     def _read_authors_from_file(cls, git_ctx) -> Tuple[str, str]:
         for file_name in cls.authors_file_names:
-            path = Path(git_ctx.repository_path / file_name)
+            path = Path(git_ctx.repository_path) / file_name
             if path.exists():
                 authors_file = path
                 break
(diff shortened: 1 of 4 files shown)


gitlint-core/gitlint/tests/contrib/rules/test_authors_commit.py
```diff
@@ -0,0 +1,88 @@
...
+from gitlint.contrib.rules.authors_commit import Authors
+
+
+Author = namedtuple("Author", "name, email")
+GitCtx = namedtuple("GitCtx", "repository_path")
+MockFile = namedtuple("MockFile", "name")
+
+
+@pytest.fixture(name="git_ctx")
+def gen_git_ctx(tmpdir):
```
jorisroovers, line 15: We should use existing fixture utility methods

gitlint-core/gitlint/tests/contrib/rules/test_authors_commit.py
```diff
@@ -0,0 +1,88 @@
...
+    return GitCtx(tmpdir)
+
+
+AUTHOR_1 = Author("John Doe", "john.doe@mail.com")
+AUTHOR_2 = Author("Bob Smith", "bob.smith@mail.com")
+
+
+@pytest.mark.parametrize("author", [AUTHOR_1, AUTHOR_2])
+def test_authors_succeeds(author, git_ctx):
+    assert not validate_authors_rule(author.name, author.email, git_ctx)
```
jorisroovers, line 27: We should match existing test styling. While correct, we should stay consistent in the project and not use pytest style tests.

gitlint-core/gitlint/tests/contrib/rules/test_authors_commit.py
```diff
@@ -0,0 +1,88 @@
+from collections import namedtuple
```
jorisroovers, line 1: Thanks for taking the time to write tests 👍 

We should however match existing test styling. While there's obviously nothing wrong with writing pytest-style tests (I do so myself in other projects), we should stay consistent within the project and not do so here. Even if we were to agree that pytest-style tests are objectively better (a debate in itself), for long-term maintainability, consistency is more important. Thanks for understanding.


This includes not using `@pytest.mark.parametrize`, and not using `assert` statements.
I recommend looking at the unit test of an existing contrib rule:

gitlint-core/gitlint/tests/contrib/rules/test_authors_commit.py
```diff
@@ -0,0 +1,88 @@
...
+from unittest.mock import Mock, patch
+
+import pytest
```
jorisroovers, line 4: Easiest is to not import pytest at all, that way we don't rely on pytest specific methods.

gitlint-core/gitlint/contrib/rules/authors_commit.py
```diff
@@ -0,0 +1,48 @@
...
+    # A rule MUST have a human friendly name
+    name = "contrib-authors-commit"
+
+    # A rule MUST have a *unique* id, we recommend starting with UC (for User-defined Commit-rule).
+    id = "CC3"
+
+    @classmethod
+    def _read_authors_from_file(cls, git_ctx) -> Tuple[str, str]:
+        for file_name in cls.authors_file_names:
+            path = Path(git_ctx.repository_path / file_name)
```
jorisroovers, line 24:
```python
File "/workspaces/gitlint/gitlint-core/gitlint/contrib/rules/authors_commit.py", line 37, in validate
    registered_authors, authors_file_name = Authors._read_authors_from_file(commit.message.context)
  File "/workspaces/gitlint/gitlint-core/gitlint/contrib/rules/authors_commit.py", line 24, in _read_authors_from_file
    path = Path(git_ctx.repository_path / file_name)
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

jorisroovers (changes requested):
Thanks for this PR, I think this makes sense to add as a contrib rule!

Notes:
- Hit an exception when trying this out, see inline, should be easily fixed
- Testing style would need updating to match existing project test style (i.e. don't import pytest), more details inline
- Missing docs, would need to be added to https://github.com/jorisroovers/gitlint/blob/main/docs/contrib_rules.md


Appreciate your contribution!

gitlint-core/gitlint/tests/contrib/rules/test_authors_commit.py
s1e2b3i4: Thanks for your feedback :+1:

My bad. I created this rule as part of another project, and I didn't think about the usage of `pytest`.
I agree with you that consistency is important.
I hope that the tests now fit the style of the other tests better.

Thanks for your great guidance!

stauchert: Thanks for your feedback :+1:

My bad. I created this rule as part of another project, and I didn't think about the usage of `pytest`.
I agree with you that consistency is important.
I hope that the tests now fit the style of the other tests better.

Thanks for your great guidance!

gitlint-core/gitlint/contrib/rules/authors_commit.py
stauchert: This should now be fixed.

jorisroovers (changes requested):
This looks really good!

2 nitpicks (also see inline):
- Shall we call this rule `contrib-allowed-authors`? I think that's more descriptive and similar to other gitlint rule names. If so, gentle reminder to also update the docs (easily forgotten 😄 )
- 1 remaining case of using `assert` over `assertEquals`

Should be straight merge after this.

stauchert:
Thanks for the note regarding the `assert` statement and the name suggestion! I agree with you on that. :+1:

jorisroovers:
Merged, super thanks!

PR #373, opened 2022-11-21
gitlint: ignore dependabot Co-authored-by lines

jorisroovers:
Dependabot tends to generate "Co-authored-by" lines that exceed the default
80 char limit. Ignore those lines in commit messages.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,7 @@
+[general]
+# See https://jorisroovers.com/gitlint/configuration/#regex-style-search
+regex-style-search=True
+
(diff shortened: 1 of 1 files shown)


jorisroovers:
Need to merge #374 first and then rebase this

webknjaz:
Talking about `Co-Authored-By:`, it could be useful for `M1: author-valid-email` to also check those trailers and the committer? You can extract these via something like `git show --no-patch --pretty='format:%(trailers:key=Co-Authored-By,separator=%x0A)%x0A%an <%ae>%x0A%cn <%ce>' HEAD`.

PR #374, opened 2022-11-21
Run tests in /tmp

jorisroovers:
run tests in /tmp to avoid gitlint picking up the .gitlint config

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -40,7 +40,8 @@ jobs:
           pip install -r test-requirements.txt
 
       - name: Unit Tests
-        run: ./run_tests.sh
+        run: ${GITHUB_WORKSPACE}/run_tests.sh ${GITHUB_WORKSPACE}/gitlint-core
+        working-directory: /tmp  # run tests in /tmp to avoid gitlint picking up the .gitlint config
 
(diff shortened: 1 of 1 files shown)


webknjaz:
FTR if you were to use pytest, it'd provide built-in capabilities to use unique tmp dirs for testing via its fixtures like `tmp_path`. This way, you could achieve isolation that spreads beyond just CI and would work the same on the contributors' machines too.

jorisroovers:
> This way, you could achieve isolation that spreads beyond just CI and would work the same on the contributors' machines too.

You're absolutely right about this, I was being hasty here. Let me re-work this. 

FWIW, we only use `pytest` to run tests today, we have no code-level dependency on it in our tests. For consistency reasons, I'm going to keep it this way - I have nothing against writing pytest-style tests (I do so in plenty of other projects), it's just that gitlint uses traditional `unittest` style tests and it will be jarring if we start mixing in `pytest` constructs at this point. If we want to start using pytest at the code level, it will require a more holistic approach. I don't think there's a strong argument for that at this time.

While I could fix this in `./run_tests.sh`, in the context of #378, that's not the right thing to do. I'll have a look at fixing this in the unit tests itself.

PR #384, opened 2022-12-01
Use Hatch for project management

jorisroovers:
Initial implementation of project management using hatch.

- Adopted workflows for builds, script running, and versioning.
- Updating contributor docs accordingly.
- Removed all setuptools related files and all requirements files.

@ofek: Appreciate if you can give this a quick review, but understand if you don't have time!

# Required follow-up

- Removing `./run_tests.sh`
    - Potentially implementing a `stats` script that replaces `./run_tests.sh --stats`, consider whether still wanted
    - Potentially implement a `clean` script that cleans up more directories that only `dist` (which is what `hatch clean` does)
- Figuring out versioning in context of pushing dev builds to pypi on every commit. Potentially using [hatch-vcs](https://github.com/ofek/hatch-vcs), but need to figure out if we can make that compatible with the need for gitlint to read the version from a file to display when doing `gitlint --version`
- Looking into workspaces once implemented in hatch:
    - https://github.com/pypa/hatch/issues/233
    - https://github.com/pypa/hatch/issues/588
- Cleanup of GHA `checks.yaml`, deduplicate between Linux and Windows. The CI script isn’t really the focus of this PR.

# Questions

CC: @ofek

- I love hatch - thanks for maintaining it! These are all just minor things.
- I noticed the `gitlint-core` packages have the dash replaced with underscore, e.g.: `gitlint_core-0.19.0.dev2.tar.gz` . AFAIK, setuptools keeps the dash. Is this just a build-tool implementation detail or is there some PEP or other rationale I’m not aware of? Just trying to learn :)
- Why is the `.gitignore` included in the build? I’ve tried adding it to the excluded files and also tried setting `ignore-vcs = true` under `[tool.hatch.build]` but that didn’t work.
- Is there a way to have a script that can run across different environments? AFAICT, there is not. Perhaps this will be supported by workspaces? Example:
    
    ```bash
    # the 'all' script runs various scripts from other environments
    [tool.hatch.envs.scripts]
    all = [
       "test:unit-tests",
       "test:format",
       "qa:integration-tests",
       "docs:build",
    ]
    ```
    
- Maybe I missed it, but from the docs, it’s not entirely clear what Hatchling is, other than a dependency of Hatch, although there’s a few references to it here and there. From my 30 sec of digging it seems the be PEP517-compliant build system that’s part of Hatch? Perhaps a short Q/A can be added to the FAQ?


ofek:
> I noticed the `gitlint-core` packages have the dash replaced with underscore, e.g.: `gitlint_core-0.19.0.dev2.tar.gz` . AFAIK, setuptools keeps the dash. Is this just a build-tool implementation detail or is there some PEP or other rationale I’m not aware of?

https://peps.python.org/pep-0625/ though you can disable `strict-naming` e.g. https://hatch.pypa.io/latest/plugins/builder/wheel/#options

> Why is the `.gitignore` included in the build?

https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection

so building from sdist is reproducible

> Is there a way to have a script that can run across different environments?

not yet

> Maybe I missed it, but from the docs, it’s not entirely clear what Hatchling is [...] Perhaps a short Q/A can be added to the FAQ?

https://github.com/pypa/hatch/commit/497408385f54e2f99a06a4cb61c3e604f0185deb

.github/workflows/checks.yml
```diff
@@ -27,39 +27,58 @@ jobs:
...
+          hatch run qa:integration-tests
         env:
           GITLINT_USE_SH_LIB: 1
-        run: ./run_tests.sh -i
 
-      - name: Code formatting (black)
-        run: ./run_tests.sh -f
+      - name: Build test (gitlint)
+        run: |
+          hatch build
```
ofek, line 55: There are cleaning flags

pyproject.toml
```diff
@@ -1,3 +1,142 @@
...
+# Environments #########################################################################################################
+# NOTE: By default all environments inherit from the 'default' environment
+
+# DEV
+# Workaround for editable install:
+# https://github.com/pypa/hatch/issues/588
+[tool.hatch.envs.dev]
+description = """
+Dev environment (running gitlint itself from source)
+"""
```
ofek, line 65: Ha, I forgot I added the ability to describe environments! I should start doing that

ofek (approved):
LGTM 🔥

jorisroovers:
For my own future reference:

>> I noticed the gitlint-core packages have the dash replaced with underscore, e.g.: gitlint_core-0.19.0.dev2.tar.gz . AFAIK, setuptools keeps the dash. Is this just a build-tool implementation detail or is there some PEP or other rationale I’m not aware of?
>
> https://peps.python.org/pep-0625/ though you can disable strict-naming e.g. https://hatch.pypa.io/latest/plugins/builder/wheel/#options

From PEP-625 I found https://packaging.python.org/en/latest/specifications/binary-distribution-format/ which specifies

> In distribution names, any run of -_. characters (HYPHEN-MINUS, LOW LINE and FULL STOP) should be replaced with _ (LOW LINE), and uppercase characters should be replaced with corresponding lowercase ones. This is equivalent to [PEP 503](https://peps.python.org/pep-0503/) normalisation followed by replacing - with _.

PR #394, opened 2022-12-19
WIP: remove dependency on sh for integration tests

jorisroovers:
Mostly involves fixing the hook tests.


jorisroovers:
I'm a bit stuck on this PR, having spent multiple hours on this without finding the solution.

Long story short, I can't get the input prompt answering to work via stdin. 

I've merged a more independent part of this work in #397 and will let this PR rest for a bit.

I believe the next steps here now are to write a very simple test program that prompts the user for some input and then see if I can get this subprocess code (or a simplified version of it) to work against that simple test program.

PR #412, opened 2023-01-16
cli: Allow for a single commit in the --commits cmd-line param

carlescufi:
In order to avoid applications needing to select between `--commit` and `--commits` based on the number of commits, allow for providing a single commit with `--commits` in the format:

```
$ gitlint --commits <sha>,
```

The trailing comma ensures that gitlint knows this is a commit list and not a refspec.

Signed-off-by: Carles Cufi <carles.cufi@nordicsemi.no>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -324,6 +324,8 @@ gitlint --commits "origin..HEAD"
 gitlint --commits 019cf40,c50eb150,d6bc75a
 # These can include special references as well
 gitlint --commits HEAD~1,mybranch-name,origin/main,d6bc75a
+# You can also lint a single commit with --commits:
+gitling --commits 019cf40,
 ```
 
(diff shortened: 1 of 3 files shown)


jorisroovers:
Thanks for the PR - makes sense 👍 

Still needs docs and a unit test (I can take care of that for you if you want).

carlescufi:
> Thanks for the PR - makes sense 👍
> 
> Still needs docs and a unit test (I can take care of that for you if you want).

Let me give it a go. If I get stuck I will let you know. Thanks!

carlescufi:
@jorisroovers done, please take another look. Thanks!

jorisroovers:
Tested locally, works well!

Merging 👍

carlescufi:
> Tested locally, works well!
> 
> Merging 👍

Thanks for the speedy review and merging!

PR #414, opened 2023-01-17
GHA: Remove duplicate windows config

jorisroovers:
Windows testing now also happens on all supported Python versions.


jorisroovers (commented):
Reword install-locall

PR #415, opened 2023-01-18
Post-release GHA testing workflow

jorisroovers:
Allows testing released gitlint versions.


jorisroovers:
Merging to main since this needs to run once on main to become available as a workflow.

Issue #380, opened 2022-11-26
[feature] Support PEP621


Freed-Wu:
According to <https://peps.python.org/pep-0621>, any python project should
store its metadata in `pyproject.toml`, not `.gitlint`, `.yamllint.yaml`, `.vintrc.yml`, `addon-info.json`, or other thing.

Many projects have support PEP621, such as

- <https://github.com/psf/black>
- <https://github.com/codespell-project/codespell>
- <https://github.com/PyCQA/pydocstyle>
- <https://github.com/pypa/setuptools>
- <https://github.com/pypa/setuptools_scm/>
- <https://github.com/pypa/hatch>
- <https://github.com/pypa/flit>
- <https://github.com/python-poetry/poetry>
- ...

So I advise this project can support, too.

Before python3.11, toml support must be provided by third-party package, such as

- <https://github.com/psf/black>: [tomli](https://github.com/psf/black/blob/main/pyproject.toml#L70)
- <https://github.com/codespell-project/codespell>: [tomli](https://github.com/codespell-project/codespell/blob/master/pyproject.toml#L46)
- <https://github.com/PyCQA/pydocstyle>: [toml](https://github.com/PyCQA/pydocstyle/blob/master/setup.py#L11)
- <https://github.com/pypa/setuptools>: [tomli, tomli-w, ini2toml](https://github.com/pypa/setuptools/blob/main/setup.cfg#L72-L81)
- <https://github.com/pypa/setuptools_scm/>: [tomli](https://github.com/pypa/setuptools_scm/blob/main/setup.cfg#L35)
- <https://github.com/pypa/hatch>: [tomli-w, tomlkit](https://github.com/pypa/hatch/blob/master/pyproject.toml#L52-L53)
- <https://github.com/pypa/flit>: [tomli](https://github.com/pypa/flit/blob/main/pyproject.toml#L32)
- <https://github.com/python-poetry/poetry> [tomli, tomlkit](https://github.com/python-poetry/poetry/blob/master/pyproject.toml#L70-L72)
- ...

You can use one as extras_require.

After python3.11, tomllib is a builtin package. Not need install any toml package. Remember `python<3.11` in requirements.txt

sigmavirus24:
> According to https://peps.python.org/pep-0621, any python project should
store its metadata in pyproject.toml, not .gitlint, .yamllint.yaml, .vintrc.yml, addon-info.json, or other thing.

[...]

Issue #115, opened 2019-12-09
Feature: pyproject.toml support for configuration


wakemaster39:
With PEP518 and pyproject.toml becoming a standard way of configuring tools I was wondering it if it something that would be considered to be supported? I am happy to see if I can make the changes necessary. I wanted to get confirmation it is something that would be accepted before I put in the effort.

sigmavirus24:
gitlint is a tool that can be used outside of the python community. Having to support only 1 configuration file (`.gitlint`) makes things easier to maintain and far less confusing for users

jorisroovers:
Thanks for your interest :-)

I agree with @sigmavirus24 on this (both from usability and maintenance perspectives).

We already have a similar request in #82, which I've left open since `tox.ini` and `setup.cfg` are `ini` files, i.e. the same as the existing gitlint configuration format. I don't think I'll get around to doing that soon though (and I haven't really investigated, there might be incompatibilities).

There's a few times I've wished I had picked something else than ini to better support complex types such as dictionaries and lists, but I don't want cause friction for end-users by switching formats at this point. The main reason I had picked ini is because that's what tools such as pylint and flake8 used at the time (just learned that pylint does support `pyproject.toml` now as well). And as pointed out, supporting multiple formats causes maintenance overhead that I prefer not to take on right now.

jorisroovers:
I won't be doing this as of now, for the aforementioned reasons.  Closing this issue.

Freed-Wu:
> gitlint is a tool that can be used outside of the python community.

An example is a codespell, which is a spell check written by python, is also a tool that can be used outside of the python community. They have added support to pyproject.toml and still keep support for `.codespellrc` and `setup.cfg`.

>  Having to support only 1 configuration file (.gitlint) makes things easier to maintain and far less confusing for users

I agree. However, if it don't support any configuration file, it will be eaier. So is its configuration redundant? I don't think so. In some aspects, user's experience is more important. Consider two situations:

One: every tool has its configuration. flake8 use `.flake8`, codespell use  `.codespellrc`, gitlint use `.gitlint`, etc.

Two: flake8 use  `[tool.flake8]` in pyproject.toml,  codespell use `[tool.codespell]` in pyproject.toml, gitlint use `[tool.gitlint]`, etc.

[...]

Issue #82, opened 2019-02-18
Read configuration from setup.cfg and tox.ini?


barbossa:
gitlint can read configuration options from `.gitlint`. Is it possible add feature for reading options from `setup.cfg` and `tox.ini`?

For example flake8 can read configuration options from `.flake8`, `setup.cfg`, or `tox.ini` files. It is very convenient

jorisroovers:
Yes, I like this idea. I'll mark this as a feature enhancement - no immediate ETA (probably a while).

jorisroovers:
I have changed my mind on this and mostly agree with @webknjaz's and @sigmavirus24 assessment at this point.
  
Supporting extra formats is a lot of maintenance overhead that I'm not willing to take on right now and it indeed can cause some complexity on the user side as well.

Having said that, with the pervasiveness of `pyproject.toml` these days, I have had some second thoughts about #115 and whether longer term we should consider supporting it. I know for gitlint itself I'm considering moving to `pyproject.toml` entirely (or as much as possible). To be clear: I mean **only** supporting `pyproject.toml` in addition to `.gitlint(.ini)`. No support for `tox.ini` or `setup.cfg` as per the reasons stated.

I think the preference of having everything sit in `pyproject.toml` or keep them all in separate config files is exactly that: preference. I can see that the occasional python developer or small project maintainer might be more inclined towards the single file for everything approach to declutter the project root - it's definitely something that speaks to me. For multi-project maintainers with more standardized and automated setups, I can relate to @webknjaz's arguments. I've worn both hats and again, think it's mostly about preference and context.

webknjaz:
@jorisroovers I think if you'll be adding TOML parsing, extending the lookup to also check for `.gitlint.toml` wouldn't be too much of a burden. But maybe it should happen at a point where Gitlint will want to support more complex configuration that INI just can't cater sanely.

sigmavirus24:
Personally, gitlint isn't a project specific to python. To the best of my knowledge people use it who aren't using it on python projects. Switching to TOML might be nice but it's not as common as a format as would be nice if we're supporting a new format. Further supporting pyproject.toml only helps python projects, not anyone else. It then introduces a filename as an option that makes no sense to anyone outside the python community. I'm not fond of that, personally. Gitlint is written in python but it's not just for python people

jorisroovers:
[...]

PR #432, opened 2023-02-08
Config: read from `setup.cfg` too, with `tool:gitlint:` section prefix

scop:
Allows consolidating config in Python projects making use of that file.

`pyproject.toml` would be kind of nicer these days, but there's no TOML parser dependency in gitlint at the moment, and adding one in order to support it would sound a bit much. It can always be added in addition to this later if desired.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -168,6 +168,17 @@ ignore=T1,body-min-length
 types = bugfix,user-story,epic
 ```
 
+## The setup.cfg file
+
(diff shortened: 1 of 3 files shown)


sigmavirus24:
For the same reasons I explained in https://github.com/jorisroovers/gitlint/issues/380 I'm a strong -1 on this

jorisroovers:
This has come up multiple times before, see #82 and #115 - we've decided not to do this at this time. 

Thanks for understanding!

scop:
Wow.

PR #457, opened 2023-03-07
GHA: github release workflow

jorisroovers:
A new GHA workflow that publishes gitlint when new gitlint releases are
created on github.


jorisroovers:
Workflows must be present on the default branch (main) to be tested/triggered. Merging this, additional commits might follow to make fixes and enhancements.

Issue #460, opened 2023-03-08
gitlint and gitlint-core v0.19.0 fail to build from sdist


jorisroovers:
This issue was first discovered and discussed in #367.

To reproduce for `gitlint-core`:

```bash
> wget https://files.pythonhosted.org/packages/29/59/c317a4a39657a4a76cd7dc0c9db4147cca57f13ebe11a469d8c6c0f88715/gitlint_core-0.19.0.tar.gz
> tar xvf gitlint_core-0.19.0.tar.gz
> cd gitlint_core-0.19.0/
> hatch build -t wheel                                                                                                                 
[wheel]
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/vscode/.local/share/hatch/env/virtual/gitlint-core/ovOkWHgP/gitlint-core-build/lib/python3.11/site-packages/hatchling/__main__.py", line 6, in <module>
    sys.exit(hatchling())
             ^^^^^^^^^^^
  File "/home/vscode/.local/share/hatch/env/virtual/gitlint-core/ovOkWHgP/gitlint-core-build/lib/python3.11/site-packages/hatchling/cli/__init__.py", line 26, in hatchling
    command(**kwargs)
  File "/home/vscode/.local/share/hatch/env/virtual/gitlint-core/ovOkWHgP/gitlint-core-build/lib/python3.11/site-packages/hatchling/cli/build/__init__.py", line 75, in build_impl
    for artifact in builder.build(
  File "/home/vscode/.local/share/hatch/env/virtual/gitlint-core/ovOkWHgP/gitlint-core-build/lib/python3.11/site-packages/hatchling/builders/plugin/interface.py", line 93, in build
    self.metadata.validate_fields()
  File "/home/vscode/.local/share/hatch/env/virtual/gitlint-core/ovOkWHgP/gitlint-core-build/lib/python3.11/site-packages/hatchling/metadata/core.py", line 243, in validate_fields
    _ = self.version
        ^^^^^^^^^^^^
  File "/home/vscode/.local/share/hatch/env/virtual/gitlint-core/ovOkWHgP/gitlint-core-build/lib/python3.11/site-packages/hatchling/metadata/core.py", line 128, in version
[...]

PR #463, opened 2023-03-09
🧪🔧 Switch to using `pypa/build` in CI

webknjaz:
This patch switches the CI/CD workflows to use the PyPA-endorsed PEP 517 build front-end called `build`. It is called without `--sdist` and `--wheel` CLI options, in which case `build` creates an sdist from the Git checkout but the following wheel build is made from that sdist tarball as opposed to using the Git checkout.

This has a side effect of smoke-testing that it's possible to build wheels from the published sdists. And this is the flow that `pip install` follows when it's requested to install from an sdist.

Ref: https://github.com/jorisroovers/gitlint/issues/460#issuecomment-1462121899

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install pypa/build
+        run: python -m pip install build==0.10.0
+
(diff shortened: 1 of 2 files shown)


sigmavirus24 (approved):

jorisroovers:
Thanks!

PR #470, opened 2023-03-10
📦 Enable support for git archive installs

### LINKED ISSUE #464: Enable support for git archive installs
See:
- https://github.com/jorisroovers/gitlint/pull/410#issuecomment-1462049126
- https://github.com/python-attrs/attrs/pull/1094#issuecomment-1408835695
- https://github.com/python-attrs/attrs/pull/1095

Suggested by @webknjaz

webknjaz:
This patch enables the end-users to succeed installing from GitHub source Git archive URLs, letting `setuptools-scm` calculate the distribution version correctly, including the exports of non-tagged commits [[1]]. `setuptools-scm` is the underlying library that `hatch-vcs` wraps, which is why the configuration pre-requisites and runtime assumptions are inherited.

Resolves #464.

[1]: https://github.com/pypa/setuptools_scm#git-archives

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Force LF line endings for text files
+* text=auto eol=lf
+
+# Needed for setuptools-scm-git-archive
+.git_archival.txt  export-subst
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,0 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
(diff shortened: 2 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Force LF line endings for text files
+* text=auto eol=lf
```
webknjaz, line 2: @jorisroovers this line/setting is not related to the PR, but it was in the file I copied and is useful in general. If you want the PR to be atomic, it can be dropped.

webknjaz:
@jorisroovers coveralls upload failed. plz restart that one job.

PR #473, opened 2023-03-21
GHA: Test Docker workflow

jorisroovers:
Allows for post-release testing of docker images.


jorisroovers:
New workflows need to be present on the main branch to be able to test them, so merging so this can be tested.

Issue #467, opened 2023-03-10
Refactor GHA pipelines for improved security


jorisroovers:
To consider as part of this:

- [ ] Ensure publishing secrets aren't available to CI jobs
- [ ] Consider adding branch protection rules
- [ ] Reconsider how dev builds are triggered from CI, splitting the it out in a separate pipeline instead of calling the workflow directly from `ci.yml`
- [ ] Other?

webknjaz:
>     * [ ]  Consider adding branch protection rules

I was referring to the environment protections, not branch protection.

>     * [ ]  Reconsider how dev builds are triggered from CI, splitting the it out in a separate pipeline instead of calling the workflow directly from `ci.yml`

I don't see a problem with this for as long as the job has a separate environment set.

webknjaz:
> [ ] Ensure publishing secrets aren't available to CI jobs

Hey @jorisroovers, I'd like to invite you to join the private beta of secretless publishing from GHA to PyPI. Please, fill out this form https://forms.gle/XUsRT8KTKy66TuUp7 to get in.

jorisroovers:
Just configured a few things:

- **Tag protection Rule:** on all tags (`*`). This effectively makes me the only one who can add or delete tags.
- **Branch Protections** on `main`:
    - **Require a pull request before merging:**
        - **Require approvals:** pull requests targeting `main` require 1 or more approvals and no changes requested before they can be merged.
        - **Require approval of the most recent reviewable push**: Whether the most recent reviewable push must be approved by someone other than the person who pushed it.
    - **Require status checks to pass before merging**: Certain status checks must pass before branches can be merged into `main`. I’ve added the Python 3.11 `tests`, `sdist-build-smoke-test`, `build-test` and `doc-checks`.
    - **Require linear history:** Prevent merge commits from being pushed to matching branches.
    - **Allow force pushes:**
        - **Specify who can force push**: @jorisroovers
- **Environment Protection Rule**:
    - **Deployment branches:** Only `main` can deploy to the `production` environment (i.e. PyPI).
    

**Notes**

- As a repo admin I can override these rules. For example, I can still do direct (force) push to main. I don’t do this often, but it happens. I’ve just tried this in 53887bc0dfb2576a2c0555d254a20bab788b7ec6.
[...]

PR #486, opened 2023-04-11
GHA: Use PyPI OIDC for publishing

jorisroovers:
Replace token based authentication with OIDC for PyPI package
publishing.

Relates to #467


webknjaz:
Wow, I didn't realize PyPI started allowing reusable workflows...

PR #491, opened 2023-05-02
Fix msg-filename kept open after edition

jorisroovers:
--msg-filename type is now a Path and not a managed click.File.
This makes it easier to control the opening and closing of the file.

Context manager will make sure that the file we use will be refreshed after
edition.

### DIFF
--- a/gitlint-core/gitlint/cli.py
+++ b/gitlint-core/gitlint/cli.py
@@ -5,6 +5,7 @@ import platform
 import stat
 import sys
 from dataclasses import dataclass
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


jorisroovers:
I ended up removing the test because I don’t think it actually tested correctly.

What we really need here is an integration test, since it seems to be OS-level behavior that’s causing the issue. However, it might be OS specific - I’ve not been able to reproduce this issue. This complicates coming up with a good integration test. Existing integration tests passing indicate there’s no major regression with gitlint itself - I think that’s good enough for now.

PR #510, opened 2023-06-23
GHA: Comment dev build numbers on PR after merge

jorisroovers:
Use GHA to leave a comment with on a PR after its merged
with a link to the corresponding dev build on PyPI and instructions
on how to install it.


jorisroovers:
The comment after this one should be from the GHA bot with the dev build number

webknjaz:
@jorisroovers should dev builds go to production PyPI? I'd argue that using TestPyPI is better for this purpose since each upload consumes disk space, and you can hit quota limits after some time, depending on the artifact sizes…
I usually only publish stable/alpha/beta/rc/post versions to PyPI, but the automatic dev only goes to TestPyPI.

For debugging the in-PR dists, I'd probably link to the GHA artifacts page for the last workflow run.

PR #539, opened 2024-02-05
Improve regexes in docs

scop:
Remove unneeded capturing group parenthesis as there's no significance to the groups, misc strictness fixes, make do the right thing with both match and search style regex matching.

### DIFF
--- a/qa/samples/config/ignore-release-commits
+++ b/qa/samples/config/ignore-release-commits
@@ -1,7 +1,7 @@
 [ignore-by-title]
-regex=^Release(.*)
+regex=^Release.*
 ignore=T5,T3
 
 [ignore-by-body]
-regex=(.*)relëase(.*)
+regex=.*relëase.*
 ignore=T3,B3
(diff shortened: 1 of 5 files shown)


qa/samples/config/ignore-release-commits
```diff
@@ -1,7 +1,7 @@
 [ignore-by-title]
-regex=^Release(.*)
+regex=^Release.*
 ignore=T5,T3
 
 [ignore-by-body]
-regex=(.*)relëase(.*)
```
scop, line 6: I'm assuming the `ë` (vs plain`e`)  was not intended here, but did not check.

jorisroovers:
Thanks for the PRs. I currently have no cycles to work on gitlint, so it will take me a while to get to them (perhaps more than a month)... But I will eventually (or @sigmavirus24)

qa/samples/config/ignore-release-commits
jorisroovers: It is intentional, to check that unicode works properly. Mostly needed back when we were supporting python 2, but there's still edge cases for which I think this is useful so I tend to include non-ascii chars in the test data.

scop: Ok, modified and rebased to keep it.

PR #563, opened 2025-08-22
Add more options for CT1

colinmarc:
This adds two options to the conventional commit contrib rule:

 - An optional whitelist of scopes (nouns)
 - An option to require a scope to be specified


colinmarc:
Looks like gha updated the default ubuntu version to one that doesn't have python 3.7.

